PR #4, opened 2021-08-02
Add code highlighting to README and simplify env creation

bryant1410:
I also changed s/source/conda/ in the command. It's the recommended way for Conda.

About removing `-f environment.yml`, note that's the default.


gabrielilharco:
Thanks for the PR!

PR #26, opened 2021-11-09
Perfom validation matrix multiply on CPU

carlini:
The current codebase computes the validation accuracy on the GPU. If you have a validation set of ~20k examples, then this means you're multiplying a 20,000 x 20,000 matrix ~= 2GB of extra GPU memory. If you're maxing out the GPU memory for training the model, the extra 2GB can make a difference. This patch just moves it to the CPU. Things are a little bit slower, but it's not much in the grand scheme of things.

### DIFF
--- a/src/training/train.py
+++ b/src/training/train.py
@@ -225,13 +225,11 @@ def evaluate(model, data, epoch, args, tb_writer=None, steps=None):
 
 def get_metrics(image_features, text_features):
     metrics = {}
-    logits_per_image = image_features @ text_features.t()
-    logits_per_text = logits_per_image.t()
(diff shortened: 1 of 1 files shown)


mitchellnw:
Thanks so much!

JunweiLiang:
```
logits_per_image = (logit_scale * image_features @ text_features.t()).detach().cpu()
```
Doesn't this still perform matmul on GPU (then put the logits_per_image on CPU)?

carlini:
Ah yes, you're right the multiply is still on the GPU. This fixed some OOM issues I was experiencing, probably because the argsort() is what was giving me problems and not the actual multiply itself. Probably though it would be better to move the entire multiply to the CPU just in case though.

PR #42, opened 2022-03-02
Merge Ross/Cade's changes onto mainline.

Vaishaal:


src/training/data.py
```diff
@@ -46,24 +52,41 @@ def __getitem__(self, idx):
...
+    if 'sizes.json' in os.listdir(dir_path):
+        sizes_filename = os.path.join(dir_path, 'sizes.json')
+        sizes = json.load(open(sizes_filename, 'r'))
+        total_size = sum(
+            [int(sizes[os.path.basename(shard)]) for shard in shards_list])
+    elif '__len__' in os.listdir(dir_path):
+        total_size = eval(open(os.path.join(dir_path, '__len__'), 'r').read())
+    else:
+        # name / path based hacks (NOTE: specific to a given download / instances of dataset in terms of samples)
+        if 'cc3m-train' in shards:
```
gabrielilharco, line 78: Can we print to the logs when this happens? It would be good to make sure the codebase is not silently doing something users don't want to. For instance, if someone is using a filtered or a bigger version of the laion dataset that contains "laion" in it, using the value 407332084 would be undesirable

src/training/data.py
```diff
@@ -118,6 +142,143 @@ def count_samples(dataloader):
...
+        self.update_env()
+        urls = self.shardsample.sample()
+        if self.epoch_shuffle:
+            if "WDS_EPOCH" not in os.environ:
+                raise ValueError(
+                    "when specifying epoch_shuffle, you must provide the epoch in the WDS_EPOCH environment variable"
+                )
+            epoch = int(os.environ["WDS_EPOCH"])
+            if self.verbose:
+                print(f"PytorchShardList epochshuffle {epoch}")
```
gabrielilharco, line 252: Can we use logging.info here instead of print? Same in other places.

src/clip/openai_clip.py
```diff
@@ -21,7 +21,7 @@
     "RN50": "https://openaipublic.azureedge.net/clip/models/afeb0e10f9e5a86da6080e35cf09123aca3b358a0c3e3b6c78a7b63bc04b6762/RN50.pt",
     "RN101": "https://openaipublic.azureedge.net/clip/models/8fa8567bab74a42d41c5915025a8e4538c3bdbe8804a470a72f30b0d94fab599/RN101.pt",
     "RN50x4": "https://openaipublic.azureedge.net/clip/models/7e526bd135e493cef0776de27d5f42653e6b4c8bf9e0f653bb11773263205fdd/RN50x4.pt",
-    "ViT-B/32": "https://openaipublic.azureedge.net/clip/models/40d365715913c9da98579312b702a82c18be219cc2a73407c4526f58eba950af/ViT-B-32.pt",
+    "ViT-B-32": "https://openaipublic.azureedge.net/clip/models/40d365715913c9da98579312b702a82c18be219cc2a73407c4526f58eba950af/ViT-B-32.pt",
```
rom1504, line 24: maybe time to update this dict from https://github.com/openai/CLIP/blob/main/clip/clip.py#L36

src/training/data.py
gabrielilharco: thanks @rwightman, adding a flag sounds good to me

src/clip/openai_clip.py
Vaishaal: Can

PR #50, opened 2022-03-29
Cleanup and refactoring for model release

rwightman:
Doing some cleanup and adding some enhanced pretrained model support. This is still in progress...

Also, switching to webdatast 0.2.x as it's installing by default from pypi now.

### DIFF
--- a/src/training/data.py
+++ b/src/training/data.py
@@ -1,30 +1,22 @@
-import os
-import sys
-import math
-import logging
-import functools
-import braceexpand
-import random
-import pdb
+import ast
 import json
-
-import pandas as pd
-import numpy as np
-import pyarrow as pa
-from PIL import Image
-
-from typing import Union
+import logging
+import math
+import os
+import random
 from dataclasses import dataclass
+from itertools import islice
 
+import braceexpand
+import numpy as np
+import pandas as pd
 import torch
+import torchvision.datasets as datasets
+import webdataset as wds
+from PIL import Image
 from torch.utils.data import Dataset, DataLoader, SubsetRandomSampler
 from torch.utils.data.distributed import DistributedSampler
-import torchvision.datasets as datasets
-from webdataset.utils import identity
-import webdataset as wds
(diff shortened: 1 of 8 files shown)


rwightman:
code is ready to be looked at, changes ended up being fairly significant as some folder renames were done to align with a good package name `open_clip`

README.md needs to be updated quite significantly still, LAION pretrain weights also need to be added but that can be done as another PR...

src/training/data.py
```diff
@@ -143,176 +131,98 @@ def count_samples(dataloader):
...
+        pipeline.extend([
+            wds.split_by_worker,
+            # at this point, we have an iterator over the shards assigned to each worker
+            wds.tarfile_to_samples(handler=log_and_continue),
+        ])
+    pipeline.extend([
+        wds.select(filter_no_caption),
+        wds.decode("pilrgb", handler=log_and_continue),
+        wds.rename(image="jpg;png", text="txt"),
+        wds.map_dict(image=preprocess_img, text=preprocess_txt),
```
rom1504, line 190: Might be good to have the handler here too

rom1504:
I looked at the code changes and it all seems pretty reasonable to me

Probably it won't impact this PR, but as an user i would appreciate from the readme:
* How to start a training on one node on a custom webdataset
* what parameters should I tune
* Same on multiple nodes
* Advises around any scheduler specifics (slurm) and distributor specific (horovod)
* how can I load a pretrained model and run an inference
* how can I fine tune
* Some loss/step curves, some zero shot acc/step curves
* advised datasets
* what is the expected training speed on hardware X
* What is the expected inference speed on hardware x

rwightman:
I'm going to merge this now so that people can start using the updated interface, new checkpoints, etc and start providing feedback. I'll be updating if issues crop up or questions arise.

I'm going to add an inference example script shortly as per a disccusion with @rom1504 , it'll also double as an inference benchmark so we test some improvements for inference speed...

I'll also work on some of the other comments re README, esp datasets

PR #58, opened 2022-04-06
Add github action to automatically push to pypi on Release x.y.z commit

rom1504:
you need to create a token there https://pypi.org/manage/account/token/
and put it in there https://github.com/mlfoundations/open_clip/settings/secrets/actions/new
name it `PYPI_PASSWORD`

The release will be triggered when you name your commit `Release x.y.z`
I advise to change the version in setup.py in that commit

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,0 +1,35 @@
+name: Release
+
+on:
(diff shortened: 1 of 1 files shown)


rom1504:
This is ready to merge after you add that secret

PR #62, opened 2022-04-08
Fix publish yml

rom1504:


rom1504:
I had removed that part by mistake
@rwightman

PR #63, opened 2022-04-09
update the dataloader

tianyu-z:
Well..I don't know why there are so many insignificant updates. Maybe it's the formater. I only updated `src/training/data.py`


tianyu-z:
A quick way to test it

```
import sys
import training.data as data
from easydict import EasyDict as edict

args = edict()
args.train_data = ["/media/work/easystore1/clotho/test_{}.tar".format(i) for i in range(16)] # all the training tar files
args.val_data = "/media/work/DATA/dataset/Clotho/train/sample1.tar" # the validation tar files
args.distributed = False
args.world_size = 1 # GPU number
args.batch_size = 6 # got issue when set larger than 1
args.workers = 3 # number of CPU for data loader
args.seed = 100 # random seed

import os
os.environ["WDS_EPOCH"] = str(10)
dl = data.get_wds_dataset(args, True).dataloader # True means train

if __name__=="__main__":
	for i, batch in enumerate(dl):
		print(i, batch)
```

tianyu-z:
Sorry, I PR to a wrong repo

PR #65, opened 2022-04-16
Gradient caching support

penfever:
Proposing integrating our implementation of gradient caching in order to enable large batch sizes on single nodes  -- this is very helpful for those without access to HPC clusters. Our implementation for OpenCLIP adapts from https://github.com/luyug/GradCache.

We have run successful tests on single-node, single GPU and single-node, multi-GPU with batch sizes up to 2048 per-GPU.

Other tweaks --

Graceful handling of PIL.UnidentifiedImageError (instead of crashing training)
Added no-eval mode for faster training loops (particularly useful for testing new features)


rwightman:
@penfever thanks for the PR, I've actually been working on an impl of grad caching locally based on the same source impl. In my experiments I found that handling the two towers separately as per the original impl allows significantly higher batch sizes (well past 4096, system memory exhaustion became the limitation instead of GPU memory), when I hacked together an impl that works with single model like yours the limit was around 2048 per GPU.

I've done single GPU testing of both schemes and they appear to be working, and I'm moving on to DDP testing now. The verification of this has been slow...

Some questions / comments;
* Do you have training logs on a standard dataset for your impl? Something like cc3m / cc12m / yfcc15m with per epoch validation numbers on a model like R50 or ViT-B-32?
* It doesn't look like you've tested with AMP as your impl isn't currently using AMP
* There's some extra code for jax impl, and function parts of grad cache that aren't used which should be removed
* `--no_eval` is not necessary, if no zero shot or eval dataset is specified no evaluation is done
* I'd like to keep logit scale in the logs, it can be a useful signal for debug

Issue #53, opened 2022-04-05
Generalizable Text Transformer Usage


Zasder3:
I've been chatting with some others interested in training CLIP for different domain tasks. They expressed interest in a simple way to use a pre-trained text transformer.

Some basic support for Hugging Face or generic classes of transformers shouldn't be too crazy of an extension to what is already fleshed out.

rwightman:
@Zasder3 that could be useful, although as far as pretrained is concerned, the LiT paper suggests it's much more useful to use pretrained + (optionally) frozen vision backbones while keeping the language trainable. However a larger variety of options for the language would be useful.

I'd proceed with a wrapper of sorts like the `TimmModel`, so say a 'HuggingFaceModel/Transformer`, keep dependencies optional, etc. We should probably start splitting timm code, hf code into separate files from the main model if we add more...

rwightman:
Some additional thoughts on this...
* `TimmModel` and related imports should move from open_clip/model.py to open_clip/timm_model.py
* create a huggingface_model.py with maybe a `HuggingFaceTransformer` (or HfTextTransformer?). This adapter module will.. 
   * create a HF model using their interface
   * add or remove any layers as needed
* Similar to the `timm_` fields in the ClipVisionCfg, we can add some `hf_` / `huggingface_` fields to ClipTextCfg which specify necessary config values for creating a hugging face text model
* create some configs with naming roughl `image_tower_name--text_tower_name.json`

The biggest mess in all of this is going to be the tokenizer. I haven't really figured out how this will work. The OpenAI CLIP tokenizer is a bit special relative to most other text transformers. It's unique in being setup to clean very ugly input. It's also a BPE tokenizer, but not quite matching the usual. A pretrained model is very much dependent on it's tokenizer, so, using the default tokenizers for pretrained models in HF we'd have to determine appropriate input cleaning, etc that works with the defaults...

rwightman:
Oh yeah, and there was somthing annoying getting in the way of the text tower encapsulation...

Some of the logic and params for the text tower are directly in the main model class right nowe main model class right now `encode_text`, etc ... which differs from the image tower that is wholly self contained.

So, we'll need a OpenClipTextTransformer for CLIP.transformer that includes this...

```    def encode_text(self, text):
        x = self.token_embedding(text)  # [batch_size, n_ctx, d_model]

        x = x + self.positional_embedding
        x = x.permute(1, 0, 2)  # NLD -> LND
[...]

PR #74, opened 2022-04-26
Gradient caching, separate text tower, more modulare file structure, more

rwightman:
* add gradient caching support via (https://github.com/luyug/GradCache/tree/main/src/grad_cache), in contrast to #65 , this impl uses the gradcache impl as is, and uses separate image / text towers via wrappers
* move text tower into `.text` submodule, will make it easier to use third party text towers (#53)  in the future, checkpoint mapping support added but more testing needed for resume, etc
* refactor files to be more granular
* training optimizer / grad cache / loss / step logic encapsulated in `TrainJig`, will also make it eaiser to support deepspeed, etc in the future

This may be the basis of a future 2.x release as it breaks checkpoint compatibility (but provides backward compat helpers). Significant grad caching testing required before being satisfied with this impl and/or #65.


rom1504:
This is old and would need to be redone. Let's close it ?

rom1504:
let's close to keep number of PR down. Can be re-opened / redone later

PR #80, opened 2022-05-04
Fix fp16

ivanprado:
The following changes are required so that the model works in fp16 mode. This PR fixes code like the following:

```py
import torch
from PIL import Image
import open_clip
import requests

device = torch.device("cuda")
model, _, preprocess = open_clip.create_model_and_transforms('ViT-B-32',
                                                             pretrained='openai',
                                                             precision="fp16",
                                                             device=device)

url = "https://raw.githubusercontent.com/mlfoundations/open_clip/main/docs/CLIP.png"
image = preprocess(Image.open(requests.get(url, stream=True).raw)).unsqueeze(0).to(device).half()
text = open_clip.tokenize(["a diagram", "a dog", "a cat"]).to(device)

with torch.no_grad():
    image_features = model.encode_image(image)
    text_features = model.encode_text(text)
    image_features /= image_features.norm(dim=-1, keepdim=True)
    text_features /= text_features.norm(dim=-1, keepdim=True)

    text_probs = (100.0 * image_features @ text_features.T).softmax(dim=-1)

print("Label probs:", text_probs)  # prints: [[1., 0., 0.]]
```

### DIFF
--- a/src/open_clip/model.py
+++ b/src/open_clip/model.py
@@ -185,7 +185,7 @@ class LayerNorm(nn.LayerNorm):
 
     def forward(self, x: torch.Tensor):
         orig_type = x.dtype
-        x = F.layer_norm(x, self.normalized_shape, self.weight, self.bias, self.eps)
+        x = F.layer_norm(x.type(torch.float32), self.normalized_shape, self.weight, self.bias, self.eps)
         return x.to(orig_type)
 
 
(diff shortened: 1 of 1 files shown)


rwightman:
@ivanprado thanks for the PR, fp16 is in a bit of an unsupported state right now, I prioritized AMP which is much better for training (and likely the same throughput on recent NVIDIA GPU + PyTorch versions for inference)... fp16 makes a difference if you're trying to fit weights into smaller memory.

That said, the OpenAI 'manual AMP' code where it casts in different spots is a bit of a hack, I'm not going to add that dtype() property back as it has multiple issues. If you ensure the input to the model is cast to float16, and cast the whole model to .half() (instead of using that convert weights fn), and leave the layernorm without the float32 cast (which I guess should have been there) it does run in pure fp16 without any float32 casts .... zero-shot results seem close

PR #85, opened 2022-05-06
Add gradient checkpointing support

rwightman:


Zasder3:
Changes are pretty minimal and all look good. Do you have any numbers for a reduction in overhead for an L/14 + the speed tradeoffs?

rwightman:
@Zasder3 it's pretty significant from testing on 1 RTX3090. For B/32 I was < 512 ... ~416 batch size w/o, and I pushed past 2048 (with 20GB of 24GB used)

PR #98, opened 2022-05-23
Adjusting SLURM section

JeniaJitsev:
Adapting SLURM section for current JUWELS Booster practices
- removing cpu-bind=none
- setting cpus-per-task=6 to fit better into current default task pinning scheme

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [[Paper]](https://arxiv.org/abs/2109.01903) [[Colab]](https://colab.research.google.com/github/mlfoundations/open_clip/blob/master/docs/Interacting_with_open_clip.ipynb)
 
-Welcome to an open source implementation of OpenAI's [CLIP](https://arxiv.org/abs/2103.00020) (Contrastive Language-Image Pre-training). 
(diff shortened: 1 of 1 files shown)


rwightman:
@JeniaJitsev why would cpus per task be 6? There are two physical cpus with 24 cores each, which gives 96 logical cores (assuming hyperthreading is enabled by default). We have 4 tasks per node (4 train processes). 

So it's either 48 / 4 = 12 (w/o hyper-threading), or 96 / 12 = 24 with...

JeniaJitsev:
@rwightman The reason is the pinning. The way default pinning works on JUWELS Booster (JWB) is such that we have 48 physical cores in JWB = 6 cores per NUMA domain (4 NUMA domains per socket). NUMA domains will be filled up with threads before the next NUMA domain will be touched => 12 threads per MPI task will be on the same NUMA domain (6 physical cores + the SMT per core). Apparently, it is good to avoid SMT, and to avoid it, we set cpus-per-task to 6. To have more cores, one would have to adapt cpu-bind manually so that the 12 threads per MPI task would be placed on the physical cores of two NUMA domains. However, 6 cores are actually good enough per GPU, so we should be just fine.

PR #108, opened 2022-06-09
Support decoupled gradient accumulation

zerovl:
Proposing the decoupled gradient accumulation~(DGA) method proposed in [the ZeroVL paper](https://arxiv.org/abs/2112.09331). DGA enables training CLIP model with a large batchsize but limited GPUs (e.g., 16,384 batchsize with 8V100 GPUs). It includes two new flags:
* --decoupled-grad-accu, a boolean flag that enables the decoupled gradient accumulation.
* --batch-size-train, a int flag that divides a large batch into several small sub-batches.

## Introduction of Decoupled Gradient Accumulation
image
The gradient of CLIP loss is decoupled into two parts. One relates to stop-grad embeddings, and the other relates to with-grad embeddings. Thus, we can mimic the large batch training by twice forward. The first forward calculates the stop-grad part, and the second forward calculates the with-grad part and backwards. The decoupled formulation helps us train CLIP with extremely large batch size. 

### Usage:
```--batch-size``` denotes the large batch size we want to use on each GPU. 
```--batch-size-train``` denotes the small batch size that one GPU can hold. 

For instance, we can separate a large batch iteration into several small batch iterations, e.g., a batch of 8,192 samples could be divided to 4 sub-batches (each has 2048 samples and could be hold with limited GPUs). For the first forward, we can obtain the stop-grad part of the large batch~(8,192 samples). For the second forward, we can forward each sub-batch, and calculate(&backward) the correct gradient of the sub-batch.

To achieve this with **8 V100 GPUs (32GB)**, ```--batch-size=1024``` will train CLIP model with 8,192 (1024 x 8GPUs) batch size. ```--batch-size-train=256``` means each iteration will be divided into 4 (1024 / 256) sub-iterations. In each sub-iteration, the small batch size is 2048 (256 x 8GPUs), which can be hold by 8 V100 GPUs.

The drawback is DGA costs more training time. In our experiments, DGA costs 30-40% extra training time.

## Experiments on Decoupled Gradient Accumulation
We prove that training w/wo DGA achieves almost the same results.

#### 1. Setting & Acc
dataset: CC3M (2.79M)
batchsize: 1024 (128 per GPU, 8 V100 32GB)
learning rate: 1e-3
weight decay: 0.1
imagenet-zeroshot-val-top1: 0.1800

#### 2. Training script
```bash
torchrun --nproc_per_node 8 -m training.main 
    --train-data "/data/cc3m/cc3m_train_anno.csv" \
    --dataset-type auto \
    --batch-size 128 \
    --precision amp \
    --workers 4 \
    --imagenet-val "/data/ILSVRC/Data/CLS-LOC/val" \
    --csv-separator , \
    --lr=1e-3 \
    --wd=0.1
```

#### 3. Log
[cc3m.log](https://github.com/mlfoundations/open_clip/files/8868295/cc3m.log)

#### 1. Setting
dataset: CC3M (2.79M)
batchsize: 1024 (128 per GPU, 8 V100 32GB)
learning rate: 1e-3
weight decay: 0.1
imagenet-zeroshot-val-top1: 0.1791

#### 2. Training script
```bash
torchrun --nproc_per_node 8 -m training.main 
    --train-data "/data/cc3m/cc3m_train_anno.csv" \
    --dataset-type auto \
    --precision amp \
    --workers 4 \
    --imagenet-val "/data/ILSVRC/Data/CLS-LOC/val" \
    --csv-separator , \
    --lr=1e-3 \
    --wd=0.1 \
    --batch-size 128 \
    --batch-size-train 64 \
    --decoupled-grad-accu
```

#### 3. Log
[cc3m+dga.log](https://github.com/mlfoundations/open_clip/files/8868296/cc3m%2Bdga.log)

### DIFF
--- a/src/open_clip/__init__.py
+++ b/src/open_clip/__init__.py
@@ -1,5 +1,5 @@
 from .factory import list_models, create_model, create_model_and_transforms, add_model_config
-from .loss import ClipLoss
+from .loss import ClipLoss, DGAClipLoss
 from .model import CLIP, CLIPTextCfg, CLIPVisionCfg, convert_weights_to_fp16, trace_model
 from .openai import load_openai_model, list_openai_models
 from .pretrained import list_pretrained, list_pretrained_tag_models, list_pretrained_model_tags,\
--- a/src/open_clip/loss.py
+++ b/src/open_clip/loss.py
@@ -1,3 +1,7 @@
+import time
+import random
+import numpy as np
+
 import torch
 import torch.distributed.nn
 from torch import distributed as dist, nn as nn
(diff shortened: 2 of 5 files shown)


rwightman:
@zerovl how does this compare to the grad caching #74 #65 impl that are being proposed? It'd be nice to do something a long the lines of the idea #74, so that we can keep one train loop, push the specifics of the impl details into a separate Loss + 'TrainJig' class

zerovl:
@rwightman the implementation details have been wrapped into a separable class. However, I am not familiar with the 'TrainJig' class. Would you check if the code works well with 'TrainJig' when you are available?

mitchellnw:
Closing for now since it's supported, thank you!

zerovl:
> Closing for now since it's supported, thank you!
@mitchellnw 
Thanks for your efforts! Would you think it is okay to mention our [eccv'22 paper](https://www.ecva.net/papers/eccv_2022/papers_ECCV/papers/136960234.pdf) about the gradient accumulation with two forwarding processes?

It would be appreciated that a part of credits goes to this paper.

mitchellnw:
Yes of course!

mitchellnw:
https://github.com/mlfoundations/open_clip/pull/281

zerovl:
@mitchellnw greatly appreciate your updating!

Issue #130, opened 2022-07-26
eot_token after truncating


liyz15:
I was comparing with the official implementation and find that 

https://github.com/mlfoundations/open_clip/blob/db0429cf256c0fae000bb00a81d40f2675d573ea/src/open_clip/tokenizer.py#L176-L178

is different from [the official implementation](https://github.com/openai/CLIP/blob/b4ae44927b78d0093b556e3ce43cbdcff422017a/clip/clip.py#L229-L232)

```python
    if len(tokens) > context_length:
        if truncate:
            tokens = tokens[:context_length]
            tokens[-1] = eot_token
```

This looks like a bug since we are using the highest id as cls token for text encoder

https://github.com/mlfoundations/open_clip/blob/c933765dc557d88e15be968e78d7580d95f86af8/src/open_clip/model.py#L451

Zasder3:
Good catch! It looks like @rom1504 fixed this in the original OpenAI repo, but the internal version didn't catch on. Will try to amend this soon.

PR #132, opened 2022-07-26
Change the last token to eot_token truncating

liyz15:
A fix for https://github.com/mlfoundations/open_clip/issues/130

### DIFF
--- a/src/open_clip/tokenizer.py
+++ b/src/open_clip/tokenizer.py
@@ -175,6 +175,7 @@ def tokenize(texts: Union[str, List[str]], context_length: int = 77) -> torch.Lo
     for i, tokens in enumerate(all_tokens):
         if len(tokens) > context_length:
             tokens = tokens[:context_length]  # Truncate
+            tokens[-1] = eot_token
         result[i, :len(tokens)] = torch.tensor(tokens)
 
     return result


rom1504:
Could you do a few inferences with the pretrained models and check if that
makes big changes ?

Ideally running the zero shot on imagenet would be great

liyz15:
Sure! Though I don't expect to see any difference, since all prompt are shorter than 77 tokens

liyz15:
I ran zero-shot on ImageNet with laion400m_e32. As expected, the modification leads little impact with an accuracy 62.92 v.s. 62.94, although in theory nothing is changed at all.

To show the difference, I changed the imagenet templates by randomly choosing 30 template and concating them together as one long template (which is longer than 77 tokens), like this
```python
random.seed(0)
random_selected_template = random.sample(openai_imagenet_template_, 30)
openai_imagenet_template = [
    lambda c: ' or '.join(template(c)[:-1] for template in random_selected_template)
]
```
With this PR, I got an accuracy of 57.49. With the original implementation I got random guess 0.001 which is expected because the sot_token becomes the largest and is used as cls_token, which sees nothing in the text.

Zasder3:
LGTM, thank you for running these tests!

Issue #99, opened 2022-05-26
Feature: support precomputed image embeddings 


rom1504:
Implementing that would make it possible to efficiently train a lit-style clip model. Which is pretty useful when a decent visual encoder is available and the goal is to map it to captions (eg multilingual ones)

https://github.com/lucidrains/DALLE2-pytorch/blob/main/train_diffusion_prior.py has a good data loader for embedding+caption, this can be reused.

We may work on this, not asking for someone else to do it :)

PR #134, opened 2022-07-27
Added ability to load pre-computed vectors from Pickle file

taylor-brown:
The "interface" matches the csv one, but with a pickled data frame.
See #99.

### DIFF
--- a/src/open_clip/model.py
+++ b/src/open_clip/model.py
@@ -251,6 +251,13 @@ class Transformer(nn.Module):
         return x
 
 
+class Precomputed(nn.Module):
+    def __init__(self,image_size=1):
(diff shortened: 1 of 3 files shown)


Zasder3:
Looks like a very helpful feature! I'll give a deeper pass over this soon. I'm the meantime, could you add an explanation of this to the `README.md`?

taylor-brown:
Sure thing. I added a commit.

rom1504:
https://github.com/rom1504/embedding-reader could be another way to get support of this
Would give : 
* reading embeddings from any FS (including S3 and gcs)
* supporting other formats (npy, parquet)

PR #150, opened 2022-08-16
add post embedding layernorm for text transformer, reportedly more st…

lucidrains:
…able by training of two large language models, Bloom and YaLM

@rom1504 one low risk intervention for your stability issues

### DIFF
--- a/src/open_clip/model.py
+++ b/src/open_clip/model.py
@@ -324,6 +324,7 @@ class CLIPTextCfg:
     width: int = 512
     heads: int = 8
     layers: int = 12
+    ln_post_emb: bool = False
 
 
 class CLIP(nn.Module):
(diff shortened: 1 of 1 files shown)


lucidrains:
also be aware that Tsinghua claims this intervention is the source of Bloom's downstream degradation https://github.com/THUDM/GLM-130B#2-embedding-layer-gradient-shrink They propose to use gradient shrinking https://github.com/mlfoundations/open_clip/pull/155 instead

lucidrains:
In other words, let's try this, and if the eval looks bad, we will consider removing it and caution future researchers to its usage. let us also consider removing https://github.com/mlfoundations/open_clip/blob/main/src/open_clip/model.py#L292 in the case we witness the degraded performance first-hand

rom1504:
We didn't end up needing this

PR #151, opened 2022-08-16
higher layernorm epsilon when doing fp16 training

lucidrains:
@rom1504 for your stability troubles

credit goes to @Veldrovive for discovering this makes a difference for Dalle2-pytorch training with fp16


rom1504:
We didn't end up needing this

PR #152, opened 2022-08-16
add stable layernorm trick from Cogview text-to-image paper, for alle…

lucidrains:
…viating overflows on the last layernorm in the sequence of attention blocks

@rom1504 again, to help with the last failed run


rom1504:
We didn't end up needing this

PR #155, opened 2022-08-21
add gradient shrinking as an option for both text and vision transfor…

lucidrains:
…mers, recommended to be set at 0.1 (10% of gradients go back to embeddings or vision patch stem)


rom1504:
We didn't end up needing this

PR #156, opened 2022-08-21
Load checkpoint on cpu, then loads into already on gpu model

rom1504:
Decrease vram usage. Neccessary for g/14

thanks @Nodja


rom1504:
I think we can merge this one @rwightman

PR #159, opened 2022-09-07
Fix: add amp_bfloat16 to precision assertion in main

iejMac:


iejMac:
@rom1504 @rwightman  small fix, amp_bfloat16 is a choice in params but it's not in the assertion.

PR #160, opened 2022-09-07
wandb: initialize name to args.name

iejMac:
seems this is strictly better:
worst case it's just as bad as random wandb name
best case you need to change nothing/less


rom1504:
@rwightman could you check this please?

rom1504:
LGTM

PR #163, opened 2022-09-15
B/32, L/14, H/14, and g/14 trained on LAION-2B. Support for huggingface_hub, allow dataset mean/std to be altered.

rwightman:


src/training/main.py
```diff
@@ -41,10 +41,13 @@ def random_seed(seed=42, rank=0):
...
-    # This enables tf32 on Ampere GPUs which is only 8% slower than
-    # float16 and almost as accurate as float32
-    # This was a default in pytorch until 1.12
-    torch.backends.cuda.matmul.allow_tf32 = True
+    if torch.cuda.is_available():
+        # This enables tf32 on Ampere GPUs which is only 8% slower than
+        # float16 and almost as accurate as float32
+        # This was a default in pytorch until 1.12
+        torch.backends.cuda.matmul.allow_tf32 = True
+        torch.backends.cudnn.benchmark = True
```
rom1504, line 49: did you observe any difference using that ?

rom1504:
lgtm

let's as well add a few lines in readme about these models ?

Issue #168, opened 2022-09-16
OpenAI CLIP -> OpenCLIP Conversion guide for H-14


trufty:
Seeing that this repo holds the new SOTA CLIP model `H-14`, a lot of people are about to begin migrating to this pip package.

I think it would be a good idea to have a conversion guide to cover some of the differences between the two packages, to make the transition easier (It was a bit painful for me yesterday). 

So I'll list the differences and issues I ran into when converting my CLIP search app:
<br>

  **Lets start with the model loader:**
  OpenAI ->
  ```py
  model, preprocess = clip.load("./models/ViT-L-14-336px.pt", device=device)
  ```
  OpenClip ->
  ```py
  model, _, preprocess = open_clip.create_model_and_transforms('ViT-H-14', device=device, pretrained="laion2b_s32b_b79k")
  ```

- You can see `create_model_and_transforms` requires an additional `pretrained` method param to download the model.  If you don't include this, your code will happily chug away on an empty model XD.
- OpenClip also returns a size 3 tuple instead of 2 in OpenAI. 
- The type of `device` param is also different.  OpenClip requires `torch.device('cuda') or torch.device('cpu')` where OpenAI can take a string of `'cuda' or 'cpu'`

<br>
<br>

**Now we can move on to image preprocessing:**
OpenAI ->
```py
image = preprocess(Image.open(path)).to(device)
```
OpenClip ->
```py
image = preprocess(Image.open(path)).unsqueeze(0).to(device)
```

- `.unsqueeze(0)` is documented in the readme here, but I just overlooked it since my OpenAI implementation does not require it.  So make sure to include it in your code
- Don't forget to push it to the correct device with `.to(device)`

<br>
<br>

**Finally we can encode the image(s):**
OpenAI ->
```py
images = torch.stack(images)
image_features = model.encode_image(images).cpu().numpy()
```
OpenClip ->
```py
image_feature = model.encode_image(image).cpu().numpy()
```

- OpenAI allows a batch of images to be encoded at once.  OpenClip is one at a time (correct me if I'm wrong)
- Lastly the returned features shape is slightly different.  OpenAI l-14 `image_features[0]` looks like `(786,)` and OpenClip H-14 `image_feature` looks like `(1, 1024)`  (I think that was correct, it's been  a day since I saw the numbers)
    - This means `image_feature.tolist()` might unpack slightly differently with OpenLClip `image_feature.tolist()[0]` was my solution
    
<br>
<br>
    
[...]

PR #169, opened 2022-09-17
Possible interface for create fn with mandatory pretrained arg and inference only transform

rwightman:
Re #168


rom1504:
Let's add it in readme then with clear instructions this should be used for pure inference, not for training

PR #176, opened 2022-09-24
Support distributed evaluation

mehdidc:
Currently, evaluation is done on rank zero. This PR provides support for distributed evaluation (using an optional `--distributed-evaluation` argument) to make evaluation faster (supports both zero-shot and retrieval).

### DIFF
--- a/src/training/data.py
+++ b/src/training/data.py
@@ -132,7 +132,7 @@ def get_imagenet(args, preprocess_fns, split):
         idxs = idxs.astype('int')
         sampler = SubsetRandomSampler(np.where(idxs)[0])
     else:
-        sampler = None
+        sampler = torch.utils.data.distributed.DistributedSampler(dataset) if args.distributed_evaluation else None
 
     dataloader = torch.utils.data.DataLoader(
         dataset,
@@ -331,6 +331,8 @@ def get_wds_dataset(args, preprocess_img, is_train, epoch=0, floor=False):
             ),
         ])
     else:
+        if args.distributed_evaluation:
+            pipeline.append(wds.split_by_node)
         pipeline.extend([
             wds.split_by_worker,
             # at this point, we have an iterator over the shards assigned to each worker
@@ -359,8 +361,10 @@ def get_wds_dataset(args, preprocess_img, is_train, epoch=0, floor=False):
         num_samples = num_batches * global_batch_size
         dataset = dataset.with_epoch(num_worker_batches)  # each worker is iterating over this
     else:
-        # last batches are partial, eval is done on single (master) node
-        num_batches = math.ceil(num_samples / args.batch_size)
+        if args.distributed_evaluation:
+            num_batches = math.ceil(num_samples / (args.batch_size * args.world_size))
+        else:
+            num_batches = math.ceil(num_samples / args.batch_size)
 
     dataloader = wds.WebLoader(
         dataset,
@@ -401,7 +405,11 @@ def get_csv_dataset(args, preprocess_fn, is_train, epoch=0):
         caption_key=args.csv_caption_key,
         sep=args.csv_separator)
     num_samples = len(dataset)
-    sampler = DistributedSampler(dataset) if args.distributed and is_train else None
+    if is_train:
+        sampler = DistributedSampler(dataset) if args.distributed else None
+    else:
+        sampler = DistributedSampler(dataset) if args.distributed_evaluation else None
+    
     shuffle = is_train and sampler is None
 
     dataloader = DataLoader(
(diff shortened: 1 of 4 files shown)


rom1504:
this LGTM

could you please double check @mitchellnw or @rwightman ?

rom1504 (approved):

mitchellnw:
In the pytorch imagenet example for distributed imagenet eval they have an `aux_val_loader` to handle the case where the test set size is not divisible by num_gpus - do we need to have this? https://github.com/pytorch/examples/blob/main/imagenet/main.py#L396-L402

src/training/train.py
```diff
@@ -232,22 +255,24 @@ def evaluate(model, data, epoch, args, tb_writer=None):
...
+    if args.wandb and is_master(args):
         assert wandb is not None, 'Please install wandb.'
         for name, val in metrics.items():
             wandb.log({f"val/{name}": val, 'epoch': epoch})
 
     return metrics
 
 
-def get_metrics(image_features, text_features, logit_scale):
+def get_metrics(image_features, text_features, logit_scale, args):
```
usuyama, line 266: nit pick but maybe better to pass a more specific argument e.g. `gather_tensor=args.distributed_evaluation`?

src/training/data.py
```diff
@@ -359,7 +361,8 @@ def get_wds_dataset(args, preprocess_img, is_train, epoch=0, floor=False):
         num_samples = num_batches * global_batch_size
         dataset = dataset.with_epoch(num_worker_batches)  # each worker is iterating over this
     else:
-        # last batches are partial, eval is done on single (master) node
+        if args.distributed_evaluation:
+            num_samples = num_samples // args.world_size
```
usuyama, line 365: maybe I'm misunderstanding, but does it skip the last few samples due to the last partial batch?

src/training/train.py
mehdidc: > In the pytorch imagenet example for distributed imagenet eval they have an `aux_val_loader` to handle the case where the test set size is not divisible by num_gpus - do we need to have this? https://github.com/pytorch/examples/blob/main/imagenet/main.py#L396-L402

Thanks for the link, not sure why they do `drop_last=True` on val_loader (not used here), probably to avoid having a GPU worker with much fewer examples than the others? so rather, they seem to do drop_last, and compute the last few examples validation performance in all GPU workers.

mehdidc: > nit pick but maybe better to pass a more specific argument e.g. `gather_tensor=args.distributed_evaluation`?

Thanks I agree, I think it's better

src/training/data.py
mehdidc: > maybe I'm misunderstanding, but does it skip the last few samples due to the last partial batch?

in evaluation, `num_samples` is only used on logging (https://github.com/mlfoundations/open_clip/blob/03839c5e00b1c21c14dbaa07921d7190a28125c5/src/training/train.py#L172, https://github.com/mlfoundations/open_clip/blob/03839c5e00b1c21c14dbaa07921d7190a28125c5/src/training/train.py#L217), it is not affecting the dataloader which depend on the wds pipeline. But it would be good to get it correct anyway, have to check how many examples  each worker exactly receive.

mehdidc:
> In the pytorch imagenet example for distributed imagenet eval they have an `aux_val_loader` to handle the case where the test set size is not divisible by num_gpus - do we need to have this? https://github.com/pytorch/examples/blob/main/imagenet/main.py#L396-L402

Thanks for the link, not sure why they do drop_last=True on val_loader (not used here), probably to avoid having a GPU worker with much fewer examples than the others? so rather, they seem to do drop_last, and compute the last few examples validation performance in all GPU workers.

mitchellnw:
@mehdidc I think this is actually necessary or else you can get different val perf when different numbers of gpus are used, e.g., see this comment: https://github.com/facebookresearch/deit/blob/main/main.py#L221-L223

mehdidc:
> @mehdidc I think this is actually necessary or else you can get different val perf when different numbers of gpus are used, e.g., see this comment: https://github.com/facebookresearch/deit/blob/main/main.py#L221-L223

I see, thanks @mitchellnw! OK so I need to fix this. I really thought that `DistributedSampler` with `drop_last=False` would do the the "right" thing (although wouldn't be ideal for disributed setting if it would be case) in the sense of seing each example exactly once (even when dataset size is not divisible by nb of workers, e.g. the remaining examples can be distributed to a subset of workers)

mehdidc:
@dmlpt Not yet, I still need to fix the val dataloader like @mitchellnw mentions

PR #178, opened 2022-09-25
Text Tower Support

iejMac:


iejMac:
@rom1504  @rwightman  This PR just moves the TextTower stuff out of the GradCache PR and updates the components to what they are in main now.

Training on 1 GPU works fine.

src/open_clip/factory.py
```diff
@@ -56,6 +56,8 @@ def load_state_dict(checkpoint_path: str, map_location='cpu'):
         state_dict = checkpoint
     if next(iter(state_dict.items()))[0].startswith('module'):
         state_dict = {k[7:]: v for k, v in state_dict.items()}
+    if 'positional_embedding' in state_dict:
```
rom1504, line 59: Why is this needed

src/open_clip/model.py
```diff
@@ -543,35 +124,17 @@ def set_grad_checkpointing(self, enable=True):
...
-        x = self.token_embedding(text)  # [batch_size, n_ctx, d_model]
-
-        x = x + self.positional_embedding
-        x = x.permute(1, 0, 2)  # NLD -> LND
-        x = self.transformer(x, attn_mask=self.attn_mask)
-        x = x.permute(1, 0, 2)  # LND -> NLD
-        x = self.ln_final(x)
+    def encode_image(self, image, normalize: bool = False):
+        features = self.visual(image)
+        return F.normalize(features, dim=-1) if normalize else features
```
rom1504, line 129: This is a breaking change as before the normalize was done outside

src/open_clip/model.py
```diff
@@ -599,6 +162,24 @@ def _convert_weights_to_fp16(l):
     model.apply(_convert_weights_to_fp16)
 
 
+def convert_state_dict(state_dict: dict):
+    if 'text_projection' in state_dict:
```
rom1504, line 166: Is this used to maintain checkpoint compatibility?
Please add a comment

src/open_clip/modified_resnet.py
```diff
@@ -0,0 +1,172 @@
+from collections import OrderedDict
```
rom1504, line 1: Is there any diff compared to code in main ?

src/open_clip/transformer.py
```diff
@@ -0,0 +1,320 @@
+from collections import OrderedDict
```
rom1504, line 1: Is there any diff compared to code in main ?

src/open_clip/model.py
rom1504: Let's undo this part (the normalize)

src/open_clip/factory.py
rom1504: Ok i see, checkpoint compatibility
Let's add a comment

iejMac: text tower change will make state dicts from before change and after change incompatible so convert_state_dict fixes that and I think
```
'positional_embeddings' in state_dict
```
just detects that. Please correct me if I'm wrong @rwightman

src/open_clip/model.py
```diff
@@ -2,414 +2,21 @@
...
 
 import numpy as np
 import torch
 import torch.nn.functional as F
 from torch import nn
 from torch.utils.checkpoint import checkpoint
 
+from .modified_resnet import ModifiedResNet
 from .timm_model import TimmModel
-from .utils import freeze_batch_norm_2d, to_2tuple
```
rom1504, line 18: Is there any diff with the moved code ?

src/open_clip/model.py
iejMac: yeah, it's also pretty nice. you think we should remove it?

rom1504: Actually ok because false by default

src/open_clip/modified_resnet.py
iejMac: updated now so all is copied from main

src/open_clip/transformer.py
iejMac: no

src/open_clip/factory.py
rom1504, line 59: Would be good to add the comment saying it's for compatibility here too

rwightman:
@iejMac this is looking good thanks, will definitely need some extensive testing before a final merge, the usage of this repo has gone up quite a bit recently

Also, further that point, in the original design I changed the one and only CLIP model class to have a separate tower, added the convert fn figuring there weren't that many users to confuse until they figured that out. With broader use now,  that's probably not the best approach. I feel we should move have two model classes now, one as is that doesn't change, and and a new one that's used for configs that enable custom text towers. We can still provide conversion fns to make it easy for one to take an existing non TT model and adapt it to a TT config for fine-tune or further train.

Thoughts? Re naming not quite sure .... ideas
* `CLIPT`
* `CustomTextCLIP`
* ...?

@rom1504 any concerns with this approach?

rom1504:
What are drawbacks to the way this PR is currently done ?

Ie why do we need 2 classes vs just changing the existing one (while keeping compatibility) ?

rwightman:
I don't see a way to keep code clean and have compatibility without forcing users to covert weights, conversion can be automatic for pretained factory use, but not if the models are being integrated non trivially into larger models, etc

rom1504:
The purpose of this PR is to make text tower support core to the lib.

That said I understand your point, probably best to keep perfect checkpoint compatibility.

About the name of the new one, could be CustomTextCLIP i guess.

rwightman:
@rom1504 by 'core feature' I meant most among the most important features. It may get there someday, or may remain a niche use case. At this point I don't think it's worth frustrating the users that have been integrating the recent large models into other systems. But do very much think it's worth adding as there are people who want to experiment with it.

Say if we trained some more checkpoints right after this merge (that don't use this feature), they would have the new state dict format, they'd be different than existing ones, lots of downstream use may require extra code / fixes, remaping old checkpoints isn't likely, having different state dict formats for models that look like they should be part of the same series/family seems off and will confuse people.

Users who've started doing things like

``` python
class MyModel(nn.Module)
  def __init__():
    self.clip = CLIP()
    self.other = Other()
```
and then use standard torch load/save state dict will be upset... etc



The custom text tower models would already stand out as different so I feel it's not an issue for them to use a different model class, the code of the model itself is fairly minimal.

rom1504:
Can you rebase on main head ? I want to see the (new) test fail

rom1504:
Looks like we need a test with direct use of the Clip class since we want to maintain compatibility to that (and it would fail here)

PR #200, opened 2022-11-05
Update README.md

mattdeitke:


mitchellnw:
in1k = ImageNet 1k

ludwigschmidt:
Should we write it out if it causes confusion?

mitchellnw:
we do in the first bullet so i think its ok

mattdeitke:
sorry, missed this!

PR #202, opened 2022-11-07
Release 2.2.0

rom1504:


rom1504:
What do you think @rwightman ?

I'd say let's release now. If we get any feedback of broken things by people we can always fix

rom1504:
Let's just do it

PR #207, opened 2022-11-07
Make HFTokenizer lazy.

rom1504:
Tokenizer is created lazily because huggingface tokenizers are not fork safe and prefer being created in each process


rom1504:
does not work, go with https://github.com/mlfoundations/open_clip/pull/208

PR #215, opened 2022-11-09
training/params.py: remove hf params and get them from model config

iejMac:


src/training/main.py
```diff
@@ -133,6 +133,11 @@ def main(args):
     )
     random_seed(args.seed, args.rank)
 
+    if hasattr(model, "text") and hasattr(model.text, "tokenizer"):
```
iejMac, line 136: not sure if it's necessary to check both. Can we expect that CustomTextCLIP will have the .tokenizer attribute?

src/training/params.py
```diff
@@ -293,13 +293,6 @@ def parse_args(args):
     parser.add_argument(
         "--grad-clip-norm", type=float, default=None, help="Gradient clip."
     )
-    # args for huggingface models support
-    parser.add_argument(
-        '--hf-model-name', type=str, default=None, help='Pretrained text encoder name from HuggingFace hub'
```
rom1504, line 298: can you also update the readme example ?

src/training/zero_shot.py
```diff
@@ -11,7 +11,8 @@
 
 
 def zero_shot_classifier(model, classnames, templates, args):
-    tokenize = HFTokenizer(args.hf_tokenizer_name) if args.hf_tokenizer_name else tokenize
+    if hasattr(model, "text") and hasattr(model.text, "tokenizer"):
```
rom1504, line 14: can we also have it at model.tokenizer that way we can let users always use model.tokenizer for normal tokenize or hf (for pretrained models) ?

rom1504: and this should be documented in readme

rom1504:
LGTM

what do you think @rwightman and @usuyama ?

rwightman:
@rom1504 I have some deadlines re paper fine-tune and other work stuff right now, but umm, puting 'non trivial' classes into models usually breaks torchscript....

rom1504:
Yeah i figure we need to provide the tokenizer to the user another way (maybe we give the user w get_tokenizer function), and pass around the tokenizer outside the model internally

usuyama:
Looking great!

Agree that putting tokenizer object into models might be problematic.

Perhaps, we can create a factory function for creating a tokenizer? e.g. `create_tokenizer(text_cfg)`
Would be great to incorporate the context_length argument from text_cfg to tokenizer as well.

rom1504:
yeah that's exactly the solution we converged on in discussions, maciej will soon give it a try

README.md
```diff
@@ -45,7 +45,7 @@ import open_clip
 model, _, preprocess = open_clip.create_model_and_transforms('ViT-B-32-quickgelu', pretrained='laion400m_e32')
 
 image = preprocess(Image.open("CLIP.png")).unsqueeze(0)
-text = open_clip.tokenize(["a diagram", "a dog", "a cat"])
+text = model.tokenizer(["a diagram", "a dog", "a cat"])
```
rom1504, line 48: Can you change this using get_tokenizer ?

src/training/zero_shot.py
```diff
@@ -4,19 +4,18 @@
...
 
 
 def zero_shot_classifier(model, classnames, templates, args):
-    tokenize = HFTokenizer(args.hf_tokenizer_name) if args.hf_tokenizer_name else tokenize
     with torch.no_grad():
         zeroshot_weights = []
         for classname in tqdm(classnames):
             texts = [template(classname) for template in templates]  # format with class
-            texts = tokenize(texts).to(args.device)  # tokenize
+            texts = model.tokenizer(texts).to(args.device)  # tokenize
```
rom1504, line 18: This should be changed to using get_tokenizet

rom1504:
Seems good now

iejMac:
nice, tests passing

rom1504:
I think it's good for merging, but will let ross take a look

rom1504 (approved):

usuyama:
Looks great. Thank you @iejMac @rom1504!

PR #218, opened 2022-11-10
Add FLAVA

gmittal:
```
python -m training.main \
    --save-frequency 1 \
    --zeroshot-frequency 1 \
    --report-to tensorboard \
    --hf-tokenizer-name gpt2 \
    --train-data=~/flava_cc/Train_GCC-training_output.csv \
    --val-data=~/flava_cc/Validation_GCC-1.1.0-Validation_output.csv \
    --csv-img-key filepath \
    --csv-caption-key title \
    --imagenet-val=~/imagenet_validation \
    --warmup 10000 \
    --batch-size=64 \
    --lr=1e-3 \
    --wd=0.1 \
    --epochs=30 \
    --workers=8 \
    --model flava-debug
```


src/training/params.py
```diff
@@ -66,6 +66,18 @@ def parse_args(args):
         default="title",
         help="For csv-like datasets, the name of the key for the captions."
     )
+    parser.add_argument(
+        "--hf-img-key",
+        type=str,
```
rom1504, line 71: What kind of hf dataset do you use ?

src/training/train.py
```diff
@@ -48,15 +48,25 @@ def train_one_epoch(model, data, epoch, optimizer, scaler, scheduler, args, tb_w
...
-    loss = ClipLoss(
-        local_loss=args.local_loss,
-        gather_with_grad=args.gather_with_grad,
-        cache_labels=True,
-        rank=args.rank,
-        world_size=args.world_size,
-        use_horovod=args.horovod)
+    if is_flava:
+        loss = FlavaLoss(
+            local_loss=args.local_loss,
```
rom1504, line 56: Does it make sense for this loss to be completely distinct? It seems flava is adding on top of clip rather than being distinct

src/training/params.py
```diff
@@ -311,7 +323,12 @@ def parse_args(args):
         action='store_true',
         help="Freeze BatchNorm running stats in image tower for any locked layers.",
     )
-
+    parser.add_argument(
+        "--context-length",
```
rom1504, line 327: This should be configured at the model type level rather than being a param

src/open_clip/flava_model.py
```diff
@@ -0,0 +1,360 @@
...
+import numpy as np
+import torch
+import torch.nn.functional as F
+from torch import nn
+
+from .transformer import Transformer
+from .utils import to_2tuple
+
+
+class VisionTransformer(nn.Module):
```
rom1504, line 14: Is it possible to instead use any visual encoder supported by openclip?

rom1504:
Left a few comments

Overall looks a bit big but interesting

Are you interested in running some experiments at scale ( for example 32 epochs of laion400m) to make sure it works ? If yes we could provide compute

rom1504:
Could you please rebase on main ?

src/training/train.py
iejMac: yeah, maybe we add all of the losses separately: MIM, MLM, ITM, MMM, GC (would be our standard ClipLoss) and as a parameter you pass in which of those you want to use. Then you could also try things like CLIP + MIM etc. And FLAVA would basically be passing in all of the possible losses

src/open_clip/flava_model.py
```diff
@@ -0,0 +1,360 @@
...
+        x = self.transformer(x)
+        x = x.permute(1, 0, 2)  # LND -> NLD
+
+        x = self.ln_post(x)
+        x = x @ self.proj
+
+        return x  # x[:, 0, :] is [CLS_I]
+
+
+class EmbeddingTransformer(nn.Module):
```
iejMac, line 72: maybe move to transformer.py

iejMac:
if we decide to go the 'merge FLAVA methods' route instead of the 'add FLAVA separately' route, we could do something where you pass parameters for what pretraining objectives you want to include and then add some ObjectiveSampler class that uses the loss functions according to some schedule. In FLAVA they use a round-robin schedule but I'd be interested to see how different ones effect training

src/training/data.py
```diff
@@ -460,11 +611,14 @@ def get_synthetic_dataset(args, preprocess_fn, is_train, epoch=0, tokenizer=None
 
     return DataInfo(dataloader, sampler)
 
+
 def get_dataset_fn(data_path, dataset_type):
     if dataset_type == "webdataset":
         return get_wds_dataset
```
usuyama, line 617: Need to pass the new collate_fn for webadaset as well? (around  `dataloader = wds.WebLoader(`)

src/training/params.py
usuyama: I agree it's better to consolidate it to model configs, but currently context_length is not reflected to the tokenizer.
Perhaps, it's a separate work item for corpora with longer text.

src/training/train.py
usuyama: I agree! It'd be awesome if we can make it modular and plug-in those losses separately.
Losses like local (token-aware) contrastive loss + MLM are important for biomedical domains too (https://arxiv.org/abs/2204.09817).

src/training/params.py
iejMac: @usuyama I made an issue for this (https://github.com/mlfoundations/open_clip/issues/222). Let's discuss how to best do it there

src/open_clip/flava_model.py
```diff
@@ -0,0 +1,360 @@
...
+            # always attend to CLS_T token
+            attend_cls = torch.zeros(x.shape[0], 1, dtype=torch.bool, device=key_padding_mask.device)
+            key_padding_mask = torch.cat([attend_cls, key_padding_mask], dim=1)
+
+        x = x.permute(1, 0, 2)  # NLD -> LND
+        x = self.transformer(x, key_padding_mask=key_padding_mask)
+        x = x.permute(1, 0, 2)  # LND -> NLD
+
+        x = self.ln_post(x)
+        x = x @ self.proj
```
apsdehal, line 123: Same as above.

src/open_clip/flava_model.py
```diff
@@ -0,0 +1,360 @@
...
+        self.itm_head = nn.Sequential(
+            nn.Linear(embed_dim, multimodal_cfg.width),
+            nn.GELU(),
+            nn.Linear(multimodal_cfg.width, 1),
+        )
+
+        self.image_to_mm_projection = nn.Linear(vision_cfg.output_dim, multimodal_cfg.width)
+        self.text_to_mm_projection = nn.Linear(text_cfg.output_dim, multimodal_cfg.width)
+
+        self.image_projection = nn.Linear(vision_cfg.output_dim, embed_dim)
```
apsdehal, line 243: Looks like you have two projections now?

src/open_clip/loss.py
```diff
@@ -119,3 +119,100 @@ def forward(self, image_features, text_features, logit_scale):
...
+        gc_loss = self.gc_loss(image_features, text_features, logit_scale)
+
+        # unimodal language
+        mlm_loss = self.mlm_loss(text_masked_recon_logits, text_masked_labels)
+
+        # multi-modal
+        itm_loss = self.itm_loss(image_text_match_logits, image_text_match_labels)
+        mm_mlm_loss = self.mlm_loss(mm_text_masked_recon_logits, text_masked_labels)
+
+        total_loss = 1.0 * gc_loss + 1.0 * mlm_loss + 1.0 * itm_loss + 1.0 * mm_mlm_loss
```
apsdehal, line 217: Make the weights configurable?

src/open_clip/tokenizer.py
```diff
@@ -187,10 +187,18 @@ def tokenize(texts: Union[str, List[str]], context_length: int = 77) -> torch.Lo
...
         self.tokenizer = AutoTokenizer.from_pretrained(tokenizer_name)
 
+        if add_pad_and_mask_tokens:
+            if self.tokenizer.pad_token_id is None:
+                assert '<|pad|>' not in self.tokenizer.get_vocab()
+                self.tokenizer.add_special_tokens({'pad_token': '<|pad|>'})
+            if self.tokenizer.mask_token_id is None:
+                assert '<|mask|>' not in self.tokenizer.get_vocab()
+                self.tokenizer.add_special_tokens({'mask_token': '<|mask|>'})
+
```
apsdehal, line 201: Original FLAVA doesn't use anything special just BERT tokenizer. Would it make sense to add original FLAVA version first and then move on to customizations?

apsdehal: Instead of doing it this way, I would suggest using HuggingFace hub to save the tokenizer with custom tokens and then loading the custom tokenizer directly. Then this custom code logic doesn't need to be added here which is anyways very specifically to experiment we will run.

src/training/data.py
```diff
@@ -30,8 +32,106 @@
...
 
+def get_collate_fn(dataset, is_flava=False):
+    if is_flava:
+        assert isinstance(dataset.tokenize, HFTokenizer)
+        hf_tokenizer = dataset.tokenize.tokenizer
+        # The following special tokens are required for
+        # the data collator to work correctly.
+        assert getattr(hf_tokenizer, 'pad_token', False)
+        assert getattr(hf_tokenizer, 'mask_token', False)
+        mlm_collator = DataCollatorForLanguageModeling(
```
apsdehal, line 43: Hmm, the original FLAVA model uses normal MLM for the text but whole word mask MLM for the multimodal MLM. We should maybe merge that later for both to use normal MLM. (Maybe leave it as the original for now).

src/training/data.py
```diff
@@ -30,8 +32,106 @@
 from open_clip.tokenizer import HFTokenizer
 
 
+def get_collate_fn(dataset, is_flava=False):
+    if is_flava:
```
apsdehal, line 36: Very specific to FLAVA logic. Maybe the collate function should take a custom preprocessing function that it applies and returns the output. Then, in your training script you can specific the custom collate function you want.

src/training/data.py
```diff
@@ -30,8 +32,106 @@
...
+                'itm': {
+                    'label': itm_labels,
+                    'image': itm_image,
+                }
+            })
+
+            # MLM
+            assert mlm_collator is not None
+            mlm_input = mlm_collator(text_list)
+            batch.update({'mlm': mlm_input})
```
apsdehal, line 81: Yeah, this definitely needs to be encapsulated. Otherwise, this will become a mess soon.

src/training/data.py
```diff
@@ -30,8 +32,106 @@
...
+
+class HFDataset(Dataset):
+    def __init__(
+        self,
+        name,
+        subset,
+        split,
+        transforms=None,
+        img_key=None,
+        caption_key=None,
```
apsdehal, line 96: Maybe make it more generic by call it text key?

src/training/data.py
```diff
@@ -30,8 +32,106 @@
...
+        tokenizer_name=None,
+        context_length=77,
```
apsdehal, line 98: Is there a typo here with tokenizer_name? You are assigning tokenize below? Maybe you want tokenizer instead everywhere?

src/training/data.py
```diff
@@ -411,13 +519,55 @@ def get_csv_dataset(args, preprocess_fn, is_train, epoch=0, tokenizer=None):
...
+        collate_fn=get_collate_fn(dataset, is_flava),
         drop_last=is_train,
     )
     dataloader.num_samples = num_samples
     dataloader.num_batches = len(dataloader)
 
     return DataInfo(dataloader, sampler)
 
+
+def get_hf_dataset(args, preprocess_fn, is_train, epoch=0, tokenizer=None):
```
apsdehal, line 531: This is returning a dataloader.

src/training/data.py
```diff
@@ -411,13 +519,55 @@ def get_csv_dataset(args, preprocess_fn, is_train, epoch=0, tokenizer=None):
...
+    dataset_str = args.train_data if is_train else args.val_data
+    identifiers = dataset_str.split(':')
+    assert 1 <= len(identifiers) <= 2, f'Invalid dataset string: {dataset_str}'
+    dataset_name = identifiers[0]
+    subset = identifiers[1] if len(identifiers) == 2 else None
+    split = 'train' if is_train else 'validation'
+    is_flava = args.model.startswith('flava')
```
apsdehal, line 538: Like mentioned above, this should be taken as preprocess function to be passed to data collator.

src/training/train.py
apsdehal: Agreed. This can be decoupled into multiple losses and then FLAVALoss using those losses as modular plugins. It would be nice to test ablations as well.

src/training/train.py
```diff
@@ -69,20 +79,60 @@ def train_one_epoch(model, data, epoch, optimizer, scaler, scheduler, args, tb_w
     end = time.time()
     for i, batch in enumerate(dataloader):
         step = num_batches_per_epoch * epoch + i
-        
+
         if not args.skip_scheduler:
             scheduler(step)
 
-        images, texts = batch
+        images, texts, text_input_masks = batch['image'], batch['text'], batch['text_input_mask']
```
apsdehal, line 86: This probably can't land as it is. We need a batch level abstraction which moves all of the elements in the batch directly to GPU when `.to('cuda')` is done. Otherwise, these kind of changes will bloat the trainer.

src/training/train.py
```diff
@@ -69,20 +79,60 @@ def train_one_epoch(model, data, epoch, optimizer, scaler, scheduler, args, tb_w
...
+            if is_flava:
+                assert 'mlm' in batch, 'FLAVA pretraining requires MLM inputs.'
+                assert 'itm' in batch, 'FLAVA pretraining requires ITM inputs.'
+
+                image_features, \
+                text_features, \
+                logit_scale, \
+                text_masked_recon, \
+                itm_logits, \
+                mm_text_masked_recon = model(
```
apsdehal, line 116: Then the batch here goes directly to the model as it is. The loss is returned and then passed down the line. This will make sure the trainer is simplified.

apsdehal (changes requested):
Thanks for this PR! the initial version is looking good. There are a lot of FLAVA specific changes which might bloat the repo if we land them as it is. 
There would be some changes required in OpenCLIP itself to make it work with a generic structure. From that point onwards, the repo will start moving towards a framework-style architecture which is also when some repos/libs/framework lose their focus and become bloated unless it is done properly. 
This will be a pivotal moment for this repo and I would though defer to the maintainers of OpenCLIP on the final decision on how they want to handle this. We can also keep this WIP as we continue developing on top.

rwightman:
Things are going to start getting out of hand w/ nature of the current train loop and adding support for different losses, needing to handle different masks, etc... 

When I was experimenting with gradient caching I decided it was best to add another layer of abstraction to the train loop. I wanted to avoid going 'full lightning' on it, but made a minimal 'TrainJig' ... could also call it a 'TrainTask' to pull optimizer setup, loss handling, task specific step logic out of the main loop.

I stopped working on grad caching because the complexity was not proving worthwhile, but might be good to bring that abstraction in now https://github.com/rwightman/open_clip/blob/grad_cache/src/training/train_jig.py

We could have a ClipTrainJig/TrainTask, FlavaTrainTask, etc... to get rid of if is_flava, etc in main loop, and allow other multi-modal train tasks to be added without further mayhem...

src/training/params.py
rwightman: also curious what hf datasets are being used? I'm not aware of any train datasets that would be fitting the FLAVA @ scale objective? Also HF Datasets needs some work before it's going to scale well to large multi-modal datasets for big cluster runs off shared storage. All of our datasets on clusters being used for training right now are in webdataset tars...

usuyama: I'm curious about this too! I tried HF dataset once for diffusion fine-tuning ([huggingface/diffusers ref](https://github.com/huggingface/diffusers/pull/356)), but it didn't really scale well + it was a bit hard to debug with Arrow format.

src/open_clip/loss.py
```diff
@@ -118,4 +118,67 @@ def forward(self, image_features, text_features, logit_scale):
...
+        vocab_size = logits.shape[-1]
+        logits = logits.reshape(-1, vocab_size)
+        labels = labels.reshape(-1)
+
+        # only compute loss on masked logits
+        masked_idx = torch.where(labels != self.ignore_index)
+        masked_logits = logits[masked_idx]
+        masked_labels = labels[masked_idx]
+
+        ce = F.cross_entropy(masked_logits, masked_labels)
```
apsdehal, line 141: Probably pass the ignore index here explicitly as well?

src/open_clip/model_configs/clip-flava-med.json
```diff
@@ -0,0 +1,17 @@
...
+  "vision_cfg": {
+      "hf_model_name": "google/vit-large-patch16-224",
+      "hf_model_pretrained": false,
+      "pooler_type": "cls_pooler",
+      "proj": "linear"
+  },
+  "text_cfg": {
+      "hf_model_name": "roberta-base",
+      "hf_model_pretrained": false,
+      "context_length": 78,
```
apsdehal, line 12: Standard is 77 I think. Are you counting CLS separately?

src/training/finetune.py
```diff
@@ -0,0 +1,234 @@
...
+
+    return dataloaders
+
+
+def compute_metrics(model, dataloader, device, args):
+    model.eval()
+    metric = get_task_metric(args.task_name)
+    val_loss = 0
+    samples_seen = 0
+    import pdb; pdb.set_trace()
```
apsdehal, line 157: Leftover? Do you want to in general check for these and make sure you don't land rouge code here?

src/training/finetune.py
```diff
@@ -0,0 +1,234 @@
+import argparse
```
apsdehal, line 1: This is specific to FLAVA. We probably shouldn't land this file into main until it is more generic.

src/training/params.py
apsdehal: @gmittal Can you reply here?

src/training/train.py
```diff
@@ -37,6 +37,26 @@ def update(self, val, n=1):
...
+        if self.keys():
+            k = list(self.keys())[0]
+            return len(self[k])
+        return 0
+
+    def to(self, *, device, non_blocking=False, dtypes={}):
+        for k, v in self.items():
+            if isinstance(v, torch.Tensor):
+                dtype = dtypes.get(k, v.dtype)
+                self[k] = v.to(device=device, non_blocking=non_blocking, dtype=dtype)
```
apsdehal, line 56: You can probably recursively call it on dicts as well. I am not sure what level of compliance you would want to go towards.

src/training/train.py
```diff
@@ -177,21 +199,22 @@ def evaluate(model, data, epoch, args, tb_writer=None):
...
                     all_image_features.append(image_features.cpu())
                     all_text_features.append(text_features.cpu())
                     logit_scale = logit_scale.mean()
                     logits_per_image = logit_scale * image_features @ text_features.t()
                     logits_per_text = logits_per_image.t()
 
-                    batch_size = images.shape[0]
+                    batch_size = batch.size()
                     labels = torch.arange(batch_size, device=device).long()
                     total_loss = (
```
apsdehal, line 219: Maybe I am missing something here. Where are you adding rest of the losses to the total loss that are returned from FLAVA?

apsdehal: or do you plan to make this particular part more generic by allowing loss registry? How are you pretraining FLAVA currently?

apsdehal (commented):
Thanks @gmittal. We are almost there. Some more comments.

PR #223, opened 2022-11-13
Add ViT-B/32 xlm roberta base.

rom1504:
First multilingual clip of openclip. Trained on full laion5B


rom1504:
initial multilingual eval done, which confirms it's working as expected. Merging

PR #228, opened 2022-11-15
Parallelize CI tests

lopho:
This PR utilizes pytest-split to split the test set based on test durations.
It also uses a cache to reuse the virtual env between runs unless `requirements*.txt` have changed. This allows skipping the setup step except on the first run, which will build the cache.

I've enabled testing for macos just to ensure that everything still works as expected with a different runner.
For this I had to add a `skipif` condition to the training test, as it was failing when pickling on MacOs.

Overall this should reduce testing times to 2-4 minutes and scale better with more tests in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
       uses: actions/cache@v3
       with:
         path: .test_durations
-        key: test_durations-${{ maxtrix.os }}-{{ matrix.job }}-${{ github.run_id }}
+        key: test_durations-${{ matrix.os }}-${{ matrix.python }}-${{ matrix.job }}-${{ github.run_id }}
         restore-keys: test_durations-0-
     - name: Setup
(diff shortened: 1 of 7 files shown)


lopho:
Note that the long run time on the first run is expected as the setup step has to be run and the durations are not available yet.
Also I noticed that currently GH is taking a long time to provide runners.

lopho:
Seems there are some issues popping up with the cache, now that it's not running on a repository owned by me.
Will have to rethink how to cache the test durations between runs.

lopho:
I used the recommended workaround to update caches. I had this before in my staging branch but tried to keep the caches a bit cleaner, this works but will pollute the actions caches with a new cache for every workflow run.
It's not bad, as they get cleared automatically when the storage limit is hit.
see: https://github.com/actions/cache/blob/main/workarounds.md

rom1504:
Is it ready now ?

PR #235, opened 2022-11-18
Option to use HF models without pretrained weights

### LINKED ISSUE #234: Use random xlm roberta / roberta in test inference
to speed up tests

cc @lopho if you feel like looking into it

lopho:
This Introduces a new parameter for `create_model`/`create_model_and_transforms` to use HF models without downloading pretrained weights: `pretrained_hf`. Defaults to `True`.

To make this change more clear, `PreTrainedTextEncoder` has been renamed to `HFTextEncoder`, which also keeps it in line with `HFTokenizer`.

It still downloads the tokenizer. ~14.2MB per model. That seems a bit large, so it may be downloading something else as well, but I am unsure what that might be.

- `test_inference_with_data` uses this option to test on random initialized HF models.
- `test_pretrained_text_encoder` still downloads a pretrained model.

fixes #234


src/open_clip/factory.py
```diff
@@ -141,9 +142,11 @@ def create_model(
                 assert False, 'pretrained image towers currently only supported for timm models'
 
         cast_dtype = get_cast_dtype(precision)
-        custom_text = model_cfg.pop('custom_text', False) or force_custom_text or ('hf_model_name' in model_cfg['text_cfg'])
+        custom_text = model_cfg.pop('custom_text', False) or force_custom_text or ('hf_model_name' in model_cfg.get('text_cfg', {}))
```
rom1504, line 145: why this change

src/open_clip/factory.py
```diff
@@ -141,9 +142,11 @@ def create_model(
...
 
         if custom_text:
+            if 'hf_model_name' in model_cfg.get('text_cfg', {}):
```
rom1504, line 148: why this change

src/open_clip/factory.py
lopho: because it could throw a KeyError if `text_cfg` is not in the config.
see https://github.com/mlfoundations/open_clip/blob/ea7718f927b84e1b46ce057d3eae5ca4c9c41434/src/open_clip/factory.py#L137
where it is done like this already.

src/open_clip/factory.py
lopho: To be sure its an huggingface model. I don't want to set unrelated parameters for model configs that aren't using HF.
Otherwise you could land in this branch if
`model_cfg['custom_text']` or `force_custom_text` is True.

rom1504:
I have a feeling this will break the models if loading them with torch.load

rom1504:
well let's go, we can always revert if indeed there is an actual problem

lopho:
@rom1504 can you give me a minimal example that could potentially break it? I'd like to test it.

PR #239, opened 2022-11-20
hf_model.py: support encoder-decoder models

iejMac:


src/open_clip/hf_configs.py
```diff
@@ -20,5 +21,16 @@
       "layers": "num_hidden_layers",
     },
     "pooler": "mean_pooler",
-  }
+  },
+  # https://huggingface.co/docs/transformers/model_doc/mt5#mt5
+  "mt5": {
+    "config_names": {
+        "context_length": "relative_attention_max_distance", # very unsure of this, couldn't find anything better
```
iejMac, line 28: couldn't find anything closer, does mt5 not have a context length?

iejMac:
I think I need to re-think the hf_configs.py file. Do we even use it? Why did we include it in the first place?

Answer that comes to mind is: so we can translate from various model configs to a common terminology for popular parameters. But if so, why don't we use those values that we include?

iejMac:
this is going to need some tokenizer tweaks too. Current version gives this error:
```
ValueError: Couldn't instantiate the backend tokenizer from one of: 
(1) a `tokenizers` library serialization file, 
(2) a slow tokenizer instance to convert or 
(3) an equivalent slow tokenizer class to instantiate and convert. 
You need to have sentencepiece installed to convert a slow tokenizer to a fast one.
```

src/open_clip/hf_model.py
```diff
@@ -86,7 +86,12 @@ def __init__(
         if config is None:
             self.config = AutoConfig.from_pretrained(model_name_or_path)
             if pretrained:
-                self.transformer = AutoModel.from_pretrained(model_name_or_path, add_pooling_layer=uses_transformer_pooler)
+                # TODO: do all model configs have this attribute? PretrainedConfig does so yes??
+                if hasattr(self.config, "is_encoder_decoder") and self.config.is_encoder_decoder:
+                    self.transformer = AutoModel.from_pretrained(model_name_or_path)
+                    self.transformer = self.transformer.encoder
```
rom1504, line 92: I don't think that's enough.
See the MT5Encoder

src/open_clip/model_configs/mt5-ViT-B-32.json
```diff
@@ -0,0 +1,16 @@
+{
```
rom1504, line 1: Can you add it to tests so we check if it works ?

rom1504:
https://pypi.org/project/sentencepiece/ maybe

src/open_clip/model_configs/mt5-ViT-B-32.json
rom1504: done

src/open_clip/hf_configs.py
rom1504: looks a bit suspicious indeed

rom1504: fixed

src/open_clip/hf_model.py
rom1504: checked, it's enough

rom1504:
thanks!

PR #242, opened 2022-11-22
hf_model.py: small refactoring

iejMac:


src/open_clip/hf_model.py
```diff
@@ -85,19 +85,15 @@ def __init__(
...
+                # self.transformer = self.transformer.encoder
+                # self.transformer.encoder = self.transformer
+                self.transformer.forward = self.transformer.encoder # TODO: check if more processing is done in forward
             else:
-                if hasattr(self.config, "is_encoder_decoder") and self.config.is_encoder_decoder:
-                    self.transformer = AutoModel.from_config(self.config)
-                    self.transformer = self.transformer.encoder
-                else:
-                    self.transformer = AutoModel.from_config(self.config, add_pooling_layer=uses_transformer_pooler)
+                self.transformer = create_func(model_args, add_pooling_layer=uses_transformer_pooler)
```
rom1504, line 96: you would need the same operation here

rom1504: ah no

PR #253, opened 2022-11-24
proper fix for hf partial freezing

rom1504:


rom1504:
@iejMac

iejMac:
nice catch, looks good

PR #256, opened 2022-11-25
adding CoCa

gpucce:
The PR idea is to add the CoCa model as implemented in https://github.com/lucidrains/CoCa-pytorch, using existing parts as much as possible.

Ideally adding possibilty to choose between custom and non custom Attention implementation as is done for CLIP.


rom1504:
It would be best to see if it's possible to unify with existing code both the tower support and the losses
So we can train with a variety of text towers and we may benefit from the existing efficient implementation of losses without too much duplicate code

gpucce:
Sure, I will try and reuse as much as I can, for now it is mostly copied from the coca-pytorch repo, will probably ask for some help while I move on :)

gpucce:
@rom1504 I will reuse the visual_model from open_clip, however in coca-pytorch the transformer layer for the text model are different from the regular ones feed_forward and attention are parallel, do you prefer like that or regular ones? I have no idea how much difference it makes.

Even if I use the regular attention I think the current implementation doesn't allow cross attention, would you prefer a CrossAttention model or adding the crossattention possibility to the regular model with kwargs to the forward?

rom1504:
I think let's bring options into current text model so they support coca
* Parallel feed forward and self attention can be added as an option. I think it won't make a huge difference for these relatively small models. Gpt-j and palm used this mecanism to improve speed at large scale.
* cross attention is indeed the important feature. I think bringing it to the existing model could be great. That should be possible for our text attention implementation. For HF encoders there's a chance it's already implemented and we can just use their implementation

Thanks for working on this!

gpucce:
@rom1504 I am moving forward, if you have time could you just have a look at how the cross attention and decoder are added to existing models to see if the integration is going in a reasonable direction?

rom1504:
Hey, posting a summary of what I see in the current PR

* coca cfg : new kind of model
* allows using txt cfg and img cfg on top of coca cfg
* coca class that uses visual and text encoders / similar to clip class
* coca loss : uses clip loss + introduces caption loss 
* adds optional cross attention in existing text transformer
* introduces MultimodalTransformerDecoder
* CoCaVisionTransformer ; why do we need this one ?
* COCaTextTransformer why do we need this one ?


I think the direction looks ok.

I would try to focus on:
* not breaking the current use cases: normal clip training and inference should still work exactly the same
* trying to minimize changes and new code as much as possible
* continuing on the integration path you took, much better than duplicating everything

thanks for the work.

@iejMac may help review as well

iejMac:
Ok before I look at it I'm putting down some of my unbiased thoughts as to how I think we should go about this:

1. The only things that we should need to add in transformers.py are an attentional pooler and a text decoder that can either be unimodal or multimodal based on what you pass into the forward function
2. A CoCa class that initializes all of it's transformers and implements the correct forward passes through each of them
3. The encoder models should work with current code - hf_model, timm_model (but this can be left for the end once everything else is working in default conditions)

Now as for the loss I see two ways of going about this:
- A specific training script and loss function for each algorithm we do (CoCa, FLAVA, EVA)
- We try to generalize a single "task" function which is parameterized by a list of loss functions you want to use so for CoCa you'd do like loss="clip,caption", for FLAVA you'd do loss="clip,MLM,MIM,..."

Not sure which one we try out first for this. Would love to hear some more thoughts on this from everyone.

iejMac:
@gpucce I think the current state of the PR is good however I see you do a lot of things to make it exactly CoCa, which is good but it creates some potentially unnecessary additions to open_clip where we could've just used something we already have in the first place. What I recommend doing is going over all of your classes and thinking about which you could just be a special case of something that already exists, then you take note of what needs to be added so it actually fits the CoCa paper and then we can discuss how to best make those additions. So for example I think AttentionPooler is a must add, but the CoCaVisionTransformer can probably be made a special case of our VisionTransformer class in transformers.py

src/open_clip/transformer.py
```diff
@@ -349,6 +472,25 @@ def forward(self, x: torch.Tensor):
 
         return x
 
+class CoCaVisionTransformer(VisionTransformer):
```
iejMac, line 475: can we make this a special case of our VisionTransformer and if there's any differences maybe we can address those in VisionTransformer nicely

src/open_clip/transformer.py
```diff
@@ -431,3 +573,102 @@ def forward(self, text):
         x = x[torch.arange(x.shape[0]), text.argmax(dim=-1)] @ self.text_projection
 
         return x
+
+class COCaTextTransformer(TextTransformer):
```
iejMac, line 577: same here, why not TextTransformer

gpucce:
@iejMac. thanks for the feedback! The reason for those `CoCa(Text/Vision)Transformer` models was that I think the forward of `(Text/Vision)Transformer` needs to be changed a bit, however after thinking about it, it should be better to redefine `embed_text` and `embed_image` for `CoCa`, in `CLIP` the same is done for `embed_text` I think, and doing so I will remove those two classes.

Instead the `AttentionPooler`, the `MultimodalTransformerDecord` and `CoCaMultimodalTransformer` should all three be needed and make sense, maybe renaming the last two just `TransformerDecoder` and `MultimodalTransformer`.  An alternative is making only one model of `TransformerDecoder` and `MultimodalTransformer` but to me it would look a bit different than the rest of the logic of `transformers.py`, however of course I'll make it as you prefer.

About point 3. in the comment above I am trying to make it like that, will see if I manage to make it work nicely with all combinations of HF and custom.

And about the training just to add what I thought about it, I think multiple script, even if they will look probably very similar, is better at first, although this may be because I don't really see the level of overlap, I don't know either EVA or FLAVA very much.

I will try and move on in the next couple of days and see where I get!

iejMac:
@gpucce so about the TransformerDecoder and MultimodalTransformer. So I agree you need the TransformerDecoder since currently there are no decoding transformers in open_clip but can't you write the TransformerDecoder such that the MultimodalTransformerDecoder is just a special case of the TransformerDecoder - one that is the result of passing in the correct initialization parameters and also inputs into forward? Maybe I'm wrong about this but it seems there are no major architectural differences between the two and it's only about what you use for qkv.

I've thought a bit more about the training situation and I think for now (as you say) we should do a separate script and then we can maybe do another method (EVA or FLAVA) at that point we can have a look at what all 3 look like and abstract the commonalities if it makes sense.

gpucce:
Hi @iejMac, I should have done most of it, if you can have a look it would be highly appreciated! 

The one thing I did different from what you said is in the end, for now, I didn't write a different train script because I was able to integrate it with very very little changes so I thought I would try it, if you think it is too messy anyway I make a new one!

It also looks like it runs on 8-A100 gpus single machine. However if you can confirm it somehow would be really nice.

Will need to revise everything but an opinion on things that need changing would be nice while I work on that!

src/open_clip/coca_model.py
```diff
@@ -0,0 +1,190 @@
...
+    layers: int = 12
+    dim_head: int = 64
+    heads: int = 8
+    clip_loss_weight: float = 1.0
+    caption_loss_weight: float = 2.0
+    n_queries: int = 256
+    dim_latents: int = None
+
+
+def _build_text_decoder_tower(
```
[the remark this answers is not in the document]
iejMac: if False you initialize a Transformer, if True you initialize a MultimodalTransformer (currently named as TransformerDecoder)

src/open_clip/loss.py
```diff
@@ -119,3 +119,48 @@ def forward(self, image_features, text_features, logit_scale):
             F.cross_entropy(logits_per_text, labels)
             ) / 2
         return total_loss
+
+
+class CoCaLoss(ClipLoss):
```
iejMac, line 124: not sure if inheritance is the right relationship here, would love some external opinion on this (@rom1504). Feels like it's more like CoCaLoss HAS ClipLoss and not IS a type of ClipLoss. Not super important though

src/open_clip/loss.py
```diff
@@ -119,3 +119,48 @@ def forward(self, image_features, text_features, logit_scale):
...
+
+        self.clip_loss_weight = clip_loss_weight
+        self.caption_loss_weight = caption_loss_weight
+        self.caption_loss = nn.CrossEntropyLoss(ignore_index=pad_id)
+
+    def forward(self, image_features, text_features, logits, labels, logit_scale):
+        clip_loss = super().forward(image_features, text_features, logit_scale)
+        clip_loss = self.clip_loss_weight * clip_loss
+
+        if self.world_size > 1:
```
iejMac, line 154: does the whole local_loss thing not apply here? why is this not like the corresponding part of the ClipLoss

src/training/main.py
```diff
@@ -267,7 +267,8 @@ def main(args):
         if is_master(args):
             logging.info(f'Start epoch {epoch}')
 
-        train_one_epoch(model, data, epoch, optimizer, scaler, scheduler, args, writer)
+        loss = create_loss(args)
```
iejMac, line 270: we shouldn't have to create the loss every epoch but this isn't very bad. It does raise an interesting question though - for methods like FLAVA there's this notion of task sampling where you sometimes train one loss and sometimes other loss, maybe we should think about how this mechanism would work. Maybe you wouldn't swap every epoch but more finely within the epoch. Anyway, for this maybe just check for training data and if that exists create the loss, otherwise set it to None

src/training/train.py
```diff
@@ -81,8 +74,9 @@ def train_one_epoch(model, data, epoch, optimizer, scaler, scheduler, args, tb_w
         optimizer.zero_grad()
 
         with autocast():
-            image_features, text_features, logit_scale = model(images, texts)
-            total_loss = loss(image_features, text_features, logit_scale)
+            loss_args = model(images, texts)
```
iejMac, line 77: make sure to adapt the evaluate function below the train_one_epoch function

iejMac:
Writing a summary as I review for future reviewers (context window too long for GPTReview 😄):

Most important files: coca_model.py, transformer.py, loss.py - that's where the CoCa stuff is implemented, rest is just integration with our stuff.

**CoCa Model**
transformer.py:
- adapts Attention, and ResidualAttentionBlock to support cross-attention functionality (base functionality should be the same).
- AttentionalPooler: cross attention with n_query learnable query vectors, used for task-specific attentional pooling (you want a different image embedding for the contrastive task and for the captioning task so you apply different poolers on top of your image encoder)
- TransformerDecoder: this is supposed currently implemented as a multimodal transformer decoder, the regular transformer decoder is just a standard Transformer which we already have implemented so we just don't pool the output

coca_model.py:
- adds TextDecoderCfg since our MultimodalTransformer (currently TransformerDecoder) has some different params
- creates CoCa model which can encode_image and encode_text separately however in it's forward call it merges the two generations through multimodal cross attention and outputs everything needed for training

loss.py:
- ads CoCaLoss class which inherits ClipLoss for contrastive loss computation 

**Training**
- adds create_loss func to factory.py which selects the loss you want to use during training
- in main.py you now also create your loss function and pass this into train_one_epoch (you do this every epoch)
- in train.py we store all of the model outputs in a single variable and unpack that into the loss class to calculate the loss without having to know what exact loss function we're using


My thoughts (@gpucce, @rom1504):
It's good you implemented the TransformerDecoder stuff because now we can see that we don't really need a TransformerDecoder. Instead I say we just rename your TransformerDecoder to a MultimodalTransformer and that will correctly encompass what it does - it's a Transformer that can take two different modalities. That way we can indeed just keep the self.cross_attn in it because that's what you need for multimodal attention.

As to training I like this approach, the algorithm should would regardless of what loss function you're using. I think we can try this out and maybe try to add one more method like FLAVA to see if all 3 fit well in this framework - parameterize your model/loss and the training code just goes with what those input/output.

Also maybe we should rename the model.py file to clip_model.py, would less ambiguous if we plan to merge this.

In general this is coming together very nicely, great job!

src/open_clip/transformer.py
```diff
@@ -85,9 +85,23 @@ def __init__(
         self.out_proj = nn.Linear(dim, dim)
         self.out_drop = nn.Dropout(proj_drop)
 
-    def forward(self, x, attn_mask: Optional[torch.Tensor] = None):
-        L, N, C = x.shape
-        q, k, v = F.linear(x, self.in_proj_weight, self.in_proj_bias).chunk(3, dim=-1)
+    def forward(self,
```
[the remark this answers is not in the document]
gpucce: yeah, since it is the same code and should not break aniything I added here too

src/open_clip/loss.py
gpucce: I had done like you say with with clip_loss as a CoCaLoss attribute but then I saw I would need to almost rewrite the init so I changed it this, can go back if it is better

src/open_clip/transformer.py
```diff
@@ -120,6 +134,27 @@ def forward(self, x, attn_mask: Optional[torch.Tensor] = None):
...
+            n_queries: int = 256,
+    ):
+        super().__init__()
+        self.query = nn.Parameter(torch.randn(n_queries, d_model))
+        self.attn = nn.MultiheadAttention(d_model, n_head)
+
+    def forward(self, kv: torch.Tensor):
+        kv = kv.permute(1, 0 ,2) # NLD -> LND
+        N = kv.shape[1]
+        kv = self.attn(self._repeat(self.query, N), kv, kv, need_weights=False)[0]
```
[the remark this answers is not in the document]
gpucce: the idea was that kv would a single vector the same width as q that gets passed as key and value now I changed it to make both explicitly passes as k and v

src/open_clip/loss.py
gpucce: I tried, I hope I got it right, I don't know too much about this

gpucce:
@iejMac One thing, is the idea that either both tokeniser and model are HF or neither of them are? Or is mixed allowed?

src/open_clip/loss.py
gpucce: @iejMac I am replying to all comments I think need a reply and I tag you in the things that I think are more relevant / I understand less.

After thinking a bit more, I think `local_loss` is not really needed in the captioning loss, since, if I gather `logits`  or `labels` then either I gather also the other one or I could have avoided gathering in the first place as the loss is "one by one", while for clip_loss I believe it is implemented to have a smaller matrix product in case of large batch size?

rom1504:
>   One thing, is the idea that either both tokeniser and model are HF or neither of them are? Or is mixed allowed?

indeed it should be both together

rom1504:
could you please add a section in the readme on how to train it ?
I figure adding a few configurations (for say B/32 and L/14 at least) would be nice too

rom1504:
oh I see you already added a config. would be nice to call it something like coca_b_32_roberta_base.json so it's explicit

rom1504:
ok looking at the config further, it seems very different from clip configs
could you add a config looking as much as possible like clip ViT-B/32 so we could compare coca vs clip easily ?

rom1504:
and last comment, just tried to run it and getting
> TypeError: __init__() got an unexpected keyword argument 'decoder_cfg'

gpucce:
I will do all the above, about not running, I was changing some things right now, now it should run, though it is not finished yet

gpucce:
> ok looking at the config further, it seems very different from clip configs could you add a config looking as much as possible like clip ViT-B/32 so we could compare coca vs clip easily ?

sure, is it ok if I keep the `coca_base.json` to have the reference for the smallest they train in the coca paper, and make a `coca_vit_32.json`

rom1504:
I added a test, so we will see directly in github if it runs or not

rom1504:
let's check if gradient checkpointing is working properly here, it seems the memory usage is very high for a quick run I did

rom1504:
ok yeah it's not about gradient checkpointing.
with local batch size of 1, it OOM on 80GB a100

gpucce:
> ok yeah it's not about gradient checkpointing. with local batch size of 1, it OOM on 80GB a100

I see it running with the same config as the test, https://wandb.ai/gpucce/open-clip/runs/e0zy499e, I must be doing smth wrong but can't understand what at the moment

rom1504:
ah I ran it in multi-node mode, could there be something wrong with distribution ?

gpucce:
> ah I ran it in multi-node mode, could there be something wrong with distribution ?

yeah that could indeed be I am not 100% confident about it and I can't test it, let me try to have a look, could you try a run with `local_loss=True`

gpucce:
@rom1504 and also in single node I see it running https://wandb.ai/gpucce/open-clip/runs/1s3zlgri

rom1504:
Is local loss still propagated via cli args ? If yes I was already using that

src/open_clip/loss.py
rom1504: I agree it would be best to avoid inheritance here

rom1504:
https://github.com/mlfoundations/open_clip/pull/256/files#diff-1a5488b81fc901e492e5cdbb6c1f99fd9618d19909adf226b748674dbafab318R200 looks like yes, so indeed there's an issue with memory

The gh action test failing seems to be showing the same problem of memory

gpucce:
It looks like it is a different test failing?

rom1504:
yes now the non regression inference test is failing, that's worse I'm afraid

rom1504:
still same error with latest commits

rom1504:
```
  File "/fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/autograd/__init__.py", line 173, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
  File "/fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/autograd/function.py", line 253, in apply
    return user_fn(self, *args)
  File "/fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/distributed/nn/functional.py", line 295, in backward
    _Reduce_Scatter.apply(ReduceOp.SUM, ctx.group, gx, *grad_outputs)
  File "/fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/distributed/nn/functional.py", line 271, in forward
    dist.reduce_scatter(tensor, list(input_tensor_list), op=op, group=group)
  File "/fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/distributed/distributed_c10d.py", line 2452, in reduce_scatter
    work = default_pg.reduce_scatter([output], [input_list], opts)
RuntimeError: Tensors must be contiguous
```

I wonder how we could test that without multi gpu/nodes

gpucce:
If you can try one last time, maybe I was doing a silly mistake, otherwise I will check better tomorrow.

rom1504:
```
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/fsx/rom1504/open_clip/src/training/main.py", line 325, in <module>
    main(sys.argv[1:])
  File "/fsx/rom1504/open_clip/src/training/main.py", line 273, in main
    train_one_epoch(model, data, loss, epoch, optimizer, scaler, scheduler, args, tb_writer=writer)
  File "/fsx/rom1504/open_clip/src/training/train.py", line 90, in train_one_epoch
    backward(total_loss, scaler)
  File "/fsx/rom1504/open_clip/src/training/train.py", line 50, in backward
    total_loss.backward()
  File "/fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/_tensor.py", line 396, in backward
    torch.autograd.backward(self, gradient, retain_graph, create_graph, inputs=inputs)
  File "/fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/autograd/__init__.py", line 173, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
  File "/fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/autograd/function.py", line 253, in apply
    return user_fn(self, *args)
  File "/fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/distributed/nn/functional.py", line 295, in backward
    _Reduce_Scatter.apply(ReduceOp.SUM, ctx.group, gx, *grad_outputs)
  File "/fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/distributed/nn/functional.py", line 271, in forward
    dist.reduce_scatter(tensor, list(input_tensor_list), op=op, group=group)
  File "/fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/distributed/distributed_c10d.py", line 2452, in reduce_scatter
    work = default_pg.reduce_scatter([output], [input_list], opts)
RuntimeError: Tensors must be contiguous
```

rom1504:
maybe putting a lot of `print(A.is_contiguous())` would help

rom1504:
```
text_latents True
image_latents True
image_latents True
logits True
labels True
total_loss c True
```
I'm a little confused then

rom1504:
https://discuss.pytorch.org/t/loss-backward-failure-due-to-contiguous-issue/96416

rom1504:
```
/fsx/rom1504/open_clip/src/training/train.py:87: UserWarning: Anomaly Detection has been enabled. This mode will increase the runtime and should only be enabled for debugging.
  with autograd.detect_anomaly():
/fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/autograd/__init__.py:173: UserWarning: Error detected in _AllGatherBackward. Traceback of forward call that caused the error:
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/fsx/rom1504/open_clip/src/training/main.py", line 325, in <module>
    main(sys.argv[1:])
  File "/fsx/rom1504/open_clip/src/training/main.py", line 273, in main
    train_one_epoch(model, data, loss, epoch, optimizer, scaler, scheduler, args, tb_writer=writer)
  File "/fsx/rom1504/open_clip/src/training/train.py", line 92, in train_one_epoch
    total_loss = loss(*model_out)
```

rom1504:
```
frame #1: c10::detail::torchCheckFail(char const*, char const*, unsigned int, char const*) + 0x60 (0x7fe33bdcb6a9 in /fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/lib/libc10
.so)
frame #2: <unknown function> + 0x1a50b1 (0x7fe2722840b1 in /fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/lib/libtorch_cuda_cpp.so)
frame #3: c10d::ProcessGroupNCCL::reduce_scatter(std::vector<at::Tensor, std::allocator<at::Tensor> >&, std::vector<std::vector<at::Tensor, std::allocator<at::Tensor> >, std::allocator<std:
:vector<at::Tensor, std::allocator<at::Tensor> > > >&, c10d::ReduceScatterOptions const&) + 0x37 (0x7fe272293a97 in /fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/lib/libtorc
h_cuda_cpp.so)
frame #4: <unknown function> + 0x9fda44 (0x7fe3393fca44 in /fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/lib/libtorch_python.so)
frame #5: <unknown function> + 0x36889d (0x7fe338d6789d in /fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/lib/libtorch_python.so)
frame #6: PyCFunction_Call + 0x59 (0x5f5b39 in /fsx/rom1504/open_clip/.env/bin/python)
frame #7: _PyObject_MakeTpCall + 0x296 (0x5f6706 in /fsx/rom1504/open_clip/.env/bin/python)
frame #8: /fsx/rom1504/open_clip/.env/bin/python() [0x50ba83]
frame #9: _PyEval_EvalFrameDefault + 0x57f2 (0x570b82 in /fsx/rom1504/open_clip/.env/bin/python)
frame #10: _PyEval_EvalCodeWithName + 0x26a (0x569d8a in /fsx/rom1504/open_clip/.env/bin/python)
frame #11: _PyFunction_Vectorcall + 0x393 (0x5f60c3 in /fsx/rom1504/open_clip/.env/bin/python)
frame #12: _PyEval_EvalFrameDefault + 0x1902 (0x56cc92 in /fsx/rom1504/open_clip/.env/bin/python)
frame #13: _PyEval_EvalCodeWithName + 0x26a (0x569d8a in /fsx/rom1504/open_clip/.env/bin/python)
frame #14: _PyFunction_Vectorcall + 0x393 (0x5f60c3 in /fsx/rom1504/open_clip/.env/bin/python)
download failed: s3://s-datasets/laion5b/laion2B-data/056080.tar to - [Errno 32] Broken pipe
frame #15: PyObject_CallObject + 0x57 (0x5f6eb7 in /fsx/rom1504/open_clip/.env/bin/python)
frame #16: THPFunction_apply(_object*, _object*) + 0x5f6 (0x7fe339060e26 in /fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/lib/libtorch_python.so)
frame #17: PyCFunction_Call + 0xfa (0x5f5bda in /fsx/rom1504/open_clip/.env/bin/python)
frame #18: _PyEval_EvalFrameDefault + 0x6465 (0x5717f5 in /fsx/rom1504/open_clip/.env/bin/python)
frame #19: _PyEval_EvalCodeWithName + 0x26a (0x569d8a in /fsx/rom1504/open_clip/.env/bin/python)
frame #20: _PyFunction_Vectorcall + 0x393 (0x5f60c3 in /fsx/rom1504/open_clip/.env/bin/python)
frame #21: PyObject_Call + 0x62 (0x5f52b2 in /fsx/rom1504/open_clip/.env/bin/python)
frame #22: _PyEval_EvalFrameDefault + 0x1f2c (0x56d2bc in /fsx/rom1504/open_clip/.env/bin/python)
frame #23: _PyEval_EvalCodeWithName + 0x26a (0x569d8a in /fsx/rom1504/open_clip/.env/bin/python)
frame #24: _PyFunction_Vectorcall + 0x393 (0x5f60c3 in /fsx/rom1504/open_clip/.env/bin/python)
frame #25: /fsx/rom1504/open_clip/.env/bin/python() [0x50b849]
frame #26: PyObject_CallObject + 0x57 (0x5f6eb7 in /fsx/rom1504/open_clip/.env/bin/python)
frame #27: torch::autograd::PyNode::apply(std::vector<at::Tensor, std::allocator<at::Tensor> >&&) + 0x1fb (0x7fe33905f86b in /fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/li
b/libtorch_python.so)
frame #28: <unknown function> + 0x3dafceb (0x7fe25be61ceb in /fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/lib/libtorch_cpu.so)
frame #29: torch::autograd::Engine::evaluate_function(std::shared_ptr<torch::autograd::GraphTask>&, torch::autograd::Node*, torch::autograd::InputBuffer&, std::shared_ptr<torch::autograd::R
eadyQueue> const&) + 0x1640 (0x7fe25be5b260 in /fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/lib/libtorch_cpu.so)
frame #30: torch::autograd::Engine::thread_main(std::shared_ptr<torch::autograd::GraphTask> const&) + 0x698 (0x7fe25be5bc98 in /fsx/rom1504/open_clip/.env/lib/python3.8/site-packages/torch/
lib/libtorch_cpu.so)
```

with `export TORCH_SHOW_CPP_STACKTRACES=1`

not sure if that helps..

rom1504:
@mitchellnw maybe you'd have some ideas on this?

gpucce:
> @rom1504 @gpucce will the contig issue above easily show up in a 4-8 GPU (single-node) run? I can take a look if it's an easy repro

Hi, I am still trying to figure smth out here, from what I have tried,
- setting `gather_with_grad` throws a similar `Tensors must be contiguous` error in 8xA100.
- setting `local_loss` shows that some parameters are ignored by the `optimization.step`

rwightman:
@gpucce reduce scatter is in the bwd of all_gather I believe, specifically gather with grad (in distributed.nn) so these seem relevant https://github.com/pytorch/pytorch/issues/73515
https://github.com/pytorch/pytorch/issues/75773

rom1504:
I'm still quite interested in trying this out

Figuring out a fix could be great

gpucce:
These week has been too busy but back on it tomorrow!

rwightman:
Is there a set of args / slurm script to repro? If I have spare cycles I can take a look as well but best to have an example to run to be efficient

rom1504:
@rwightman same parameters we use for clip usually (so eg laion2B 13B) except you use coca_ViT-B-32 as model type
And probably put batch size 1 to start

iejMac:
@rwightman, @gpucce, @rom1504 can we merge this into this PR? https://github.com/gpucce/open_clip/pull/1

We don't need to be gathering things for the captioning loss so I removed those things and also renamed model.py to clip_model.py

iejMac:
actually on second thought, reverted the clip_model.py renaming

gpucce:
I merged as without the renaming it just makes things better.

iejMac:
@gpucce one more improvement! https://github.com/gpucce/open_clip/pull/2

rom1504:
@iejMac started a run and it seems it's going well https://wandb.ai/iejmac/open-clip/reports/CoCa-ViT-B-32-v1--VmlldzozMTcyODc2

one important feature we're missing here is ability to do AR sampling, ie captioning.

I'd suggest to add that by looking at https://github.com/lucidrains/x-transformers/blob/main/x_transformers/autoregressive_wrapper.py

rom1504:
Before merge we'd like to:
* [ ] make sure normal clip is still working the same (#198 can help but otherwise just doing runs)
* [ ] check if we really need the change for qkv in transformer model
* [ ] check if anything need to be cleaned up
* [ ] check captioning works

Bonus features include:
* [ ] report the captioning loss and contrastive loss independently in wandb

rom1504:
would appreciate your review @rwightman if you think anything big need to be done

gpucce:
Thanks for sharing the wandb! 

I will get to work on the missing things starting from captioning as soon as possible, starting tomorrow I might be able to do some of it.

rom1504:
Another idea of bonus feature (not for this PR probably) : support many HF decoder for the "multimodal transformer" that got added here

iejMac:
@gpucce Do you think you could give me push access to your fork? I'd love to help out but I don't want you to have to manually merge all of my suggested changes each time I make them

gpucce:
> @gpucce Do you think you could give me push access to your fork? I'd love to help out but I don't want you to have to manually merge all of my suggested changes each time I make them

Sure, I will do it as soon as I am on a computer

rwightman:
> would appreciate your review @rwightman if you think anything big need to be done

Made some code review comments, most important points:
* Make CoCa model dual tower (.visual, .text) from the start so builtin and HF text transformer models are the same, no bwd compat to be concerned about
* Revert changes to Attention / CustomResidualAttentionBlock, we should avoid using them with the CoCa model
* Move all models to output dicts instead of tuples, makes it more flexible pass output through optional loss / filters w/o brittle indexing and tuple len checks that are bound to fail with one more addition. Could make the dict output optional at construction time to prevent possible breaks for other downstream users.

and then test test test.

gpucce:
Thanks for the review @rwightman. @iejMac had raised a similar point to the second one. To coordinate with everyone (@rom1504) since the list of todos is getting longer, I am planning to address all of them, however I don't proceed too fast. If someone is working on some of them to speed up the whole process, please share with me.

Otherwise I will make everything as suggested taking a bit of time, in general will start from the generative part.

rom1504:
yes I think starting with implementing captioning will give us confidence that things are working

src/open_clip/coca_model.py
```diff
@@ -0,0 +1,200 @@
...
+        super().__init__()
+
+
+        norm_layer = (
+            LayerNormFp32
+            if cast_dtype in (torch.float16, torch.bfloat16)
+            else LayerNorm
+        )
+
+        text = _build_input_dependent_text_tower(embed_dim, text_cfg, quick_gelu, cast_dtype, multimodal=False)
```
rom1504, line 86: could be self.text here

src/open_clip/coca_model.py
```diff
@@ -0,0 +1,200 @@
...
+        return (image_latent, x[:, 1:]) if return_tokens else image_latent
+
+    def _repeat(self, t, N):
+        return t.reshape(1, 1, -1).repeat(N, 1, 1)
+
+    def encode_text(self, text, normalize=True, return_tokens=False):
+        text = text[:, :-1] # make space for CLS token
+        cast_dtype = self.transformer.get_cast_dtype()
+
+        # cls_mask = (text!=self.pad_id).unsqueeze(1)
```
rom1504, line 167: is this unneeded now ?

src/open_clip/transformer.py
```diff
@@ -485,3 +558,86 @@ def forward(self, text):
...
+        mask = torch.empty(self.context_length, self.context_length)
+        mask.fill_(float("-inf"))
+        mask.triu_(1)  # zero out the lower diagonal
+        return mask
+
+    def forward(self, image_embs, text_embs):
+        text_embs = text_embs.permute(1, 0, 2)  # NLD -> LND
+        image_embs = image_embs.permute(1, 0, 2)  # NLD -> LND
+
+        for r, ca in zip(self.resblocks, self.cross_attn):
```
rom1504, line 631: could you name those resblock and cross_attn ? r and ca are a bit confusing

src/open_clip/transformer.py
```diff
@@ -485,3 +558,86 @@ def forward(self, text):
...
+
+    def build_attention_mask(self):
+        # lazily create causal attention mask, with full attention between the vision tokens
+        # pytorch uses additive attention mask; fill with -inf
+        mask = torch.empty(self.context_length, self.context_length)
+        mask.fill_(float("-inf"))
+        mask.triu_(1)  # zero out the lower diagonal
+        return mask
+
+    def forward(self, image_embs, text_embs):
```
rom1504, line 627: could you add a comment explaining the shape of those, I think it'll help

src/open_clip/coca_model.py
```diff
@@ -0,0 +1,200 @@
...
+
+        self.logit_scale = nn.Parameter(torch.ones([]) * np.log(1 / 0.07))
+
+    @torch.jit.ignore
+    def set_grad_checkpointing(self, enable=True):
+        self.visual.set_grad_checkpointing(enable)
+        self.transformer.grad_checkpointing = enable
+        self.multimodal_decoder.grad_checkpointing = enable
+
+    def encode_image(self, images, normalize=True, return_tokens=False):
```
rom1504, line 128: any chance we can make this look like this instead:
```
 def encode_image(self, images, normalize=True, return_tokens=False):
        return self.visual(images, normalize=normalize, return_tokens=return_tokens) 
 ```
 ? and add the option in visual to return tokens

File: src/open_clip/coca_model.py [hunk 15]
```diff
@@ -0,0 +1,200 @@
...
+
+        image_latent = x[:, 0]
+        image_latent = F.normalize(image_latent, dim=-1) if normalize else image_latent
+
+        return (image_latent, x[:, 1:]) if return_tokens else image_latent
+
+    def _repeat(self, t, N):
+        return t.reshape(1, 1, -1).repeat(N, 1, 1)
+
+    def encode_text(self, text, normalize=True, return_tokens=False):
```
rom1504, line 163: same comment as for visual, can we use the text tower much more ?

rom1504:
@AwePhD help would definitely be appreciated. See above comments for what we need

You can open PRs on the branch of this PR

Soonhwan-Kwon:
I'm also want to help, and I've done many experiments with CoCa model on most public datasets, and caption generation also(but w/o HF compatibility).

rom1504:
Ok. I think the best path forward is I'm going to merge this in a coca
branch in this repo, and then we all do PRs towards that branch.

When we're happy that everything is good, we merge to main

rom1504:
Recent news is we have confirmed with Maciej that:
- this gets the same perf on zero shot classification than clip B/32
- captioning seems to work a bit
We can probably share this model to help testing

We need to proper AR sampling implementation to know whether captioning is
working well.

gpucce:
Nice that it gets same performance! 

If you can wait a moment before merging in a few moments I should be able to simplify the logic for the visual part, while for the text one more things would need changing.

gpucce:
And if you can somehow share the model that would be very useful

src/open_clip/coca_model.py
```diff
@@ -160,19 +155,31 @@ def encode_image(self, images, normalize=True, return_tokens=False):
     def _repeat(self, t, N):
         return t.reshape(1, 1, -1).repeat(N, 1, 1)
 
+    def _build_cls_mask(self, text, cast_dtype):
```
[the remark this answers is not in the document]
gpucce: I think right now the cls token at the end can attend to pad tokens in the sequence, this should not be possible with this extra mask

src/open_clip/coca_model.py
gpucce: Not just one line, but now visual is reused.

src/open_clip/coca_model.py
gpucce: This one is a bit harder than the visual one I think

src/open_clip/coca_model.py
gpucce: I think adding this would make the state_dict of the model you have just trained incompatible, is that fine?

gpucce:
@rom1504 the visual part should be simpler now, can make the new branch now, I will still work on the generative part as soon as I have time

src/open_clip/coca_model.py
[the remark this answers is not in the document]
rom1504: sounds good!

src/open_clip/transformer.py
```diff
@@ -465,6 +465,9 @@ def forward(self, x: torch.Tensor):
         x = self.transformer(x)
         x = x.permute(1, 0, 2)  # LND -> NLD
 
+        if output_tokens:
+            return x
+        
         if self.global_average_pool:
             x = x.mean(dim=1)
```
rom1504, line 472: I'm wondering if this can be done after the ln post

if yes, then it will make it possible to do the ln post only here and not in coca

src/open_clip/coca_model.py
```diff
@@ -123,22 +123,7 @@ def set_grad_checkpointing(self, enable=True):
...
-            ],
-            dim=1,
-        )  # shape = [*, grid ** 2 + 1, width]
-        x = x + self.visual.positional_embedding.to(x.dtype)
-        x = self.visual.ln_pre(x)
-        x = x.permute(1, 0, 2)  # NLD -> LND
-        x = self.visual.transformer(x)
-        x = x.permute(1, 0, 2)  # LND -> NLD
+        x = self.visual(images, output_tokens=True)
         x = self.visual.ln_post(x)
```
rom1504, line 127: this ln post call makes a big assumption on the API of the visual encoder

src/open_clip/coca_model.py
rom1504: what is missing? don't we simply need to add that return tokens option in the text encoder too ?

rom1504:
@gpucce I merged into coca branch. I excluded the 2 last commits you added today to avoid discrepancies with our trained model.
Can you please open a PR with your 2 last commits and also any further improvement ? targeting `coca` branch, and not main
Thanks

All comments mentioned here stay valid

src/open_clip/coca_model.py
rom1504: yes, we'll retrain

rom1504:
Please refer to https://github.com/mlfoundations/open_clip/pull/308 for follow ups

PR #257, opened 2022-11-25
add script for only extracting clip features

RyanMarten:


rom1504:
what about using https://github.com/rom1504/clip-retrieval instead ? it has the same functionality with more options and doc

PR #259, opened 2022-11-26
fixing data filtering

gabrielilharco:
The current code ignores samples where the image ends with `jpeg`, when there is no need to. In the extreme case where all the samples have the `jpeg` extension, the current code will silently hang forever with the `--dataset-resampled` flag, as it will keep trying to find valid samples but there will be none.


rom1504:
lgtm thanks

PR #264, opened 2022-11-28
add deepspeed zero-stage-1

Quan-Sun:
Signed-off-by: Sun Quan <sunquan@baai.ac.cn>
Add deepspeed zero-stage-1 to the training code

PR #268, opened 2022-11-29
support layer decay and different lr for text/visual encoder

Quan-Sun:
Signed-off-by: quansun <sunquan@baai.ac.cn>
Add new features:
- layer decay with value depending on layer
- support for different lr for text and image

### DIFF
--- a/src/training/main.py
+++ b/src/training/main.py
@@ -31,7 +31,7 @@ from training.logger import setup_logging
 from training.params import parse_args
 from training.scheduler import cosine_lr
 from training.train import train_one_epoch, evaluate
-
+from training.optim import LayerDecayValueAssigner, create_optimizer
 
 def random_seed(seed=42, rank=0):
     torch.manual_seed(seed + rank)
@@ -133,6 +133,22 @@ def main(args):
     )
     random_seed(args.seed, args.rank)
 
+    model_without_ddp = model
+    visual_ld = args.visual_ld if args.visual_ld else args.ld
+    text_ld = args.text_ld if args.text_ld else args.ld
+    
+    if visual_ld < 1.0:
+        visual_num_layers = model_without_ddp.visual.get_num_layers()
+        assigner_visual = LayerDecayValueAssigner(list(visual_ld ** (visual_num_layers + 1 - i) for i in range(visual_num_layers + 2)))
+    else:
+        assigner_visual = None
+
+    if text_ld < 1.0:
+        text_num_layers = model_without_ddp.text.get_num_layers()
+        assigner_text = LayerDecayValueAssigner(list(text_ld ** (text_num_layers + 1 - i) for i in range(text_num_layers + 2)))
+    else:
+        assigner_text = None
+
     if args.trace:
         model = trace_model(model, batch_size=args.batch_size, device=device)
 
@@ -150,6 +166,10 @@ def main(args):
         model.set_grad_checkpointing()
 
     if is_master(args):
+        if assigner_visual is not None:
+            logging.info("Assigned visual lr values = %s" % str(assigner_visual.values))
+        if assigner_text is not None:
+            logging.info("Assigned text lr values = %s" % str(assigner_text.values))
         logging.info("Model:")
         logging.info(f"{str(model)}")
         logging.info("Params:")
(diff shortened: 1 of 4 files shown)


Quan-Sun:
Just a follow-up. Is anyone taking a look?

gabrielilharco:
Hi @Quan-Sun, thanks for the PR! Can you give more context on why this is helpful? Have you observed better results with either of these changes?

Quan-Sun:
Hello Gabriel, thanks for your reply! This PR is for layer decay and different lr for text/visual encoder. Learning rate layer decay is a common trick when we train a big model loading pre-trained weights. I think text and visual encoders are different due to natural differences between image and text, so different learning rates and values of learning rate layer decay should be applied. 
We use this to train [EVA-CLIP](https://github.com/baaivision/EVA/tree/master/clip) which is the largest performant open-sourced CLIP model evaluated via zero-shot classification performance and achieves the state-of-the-art top-1 accuracy on ImageNet-1K among all self-supervised learning approaches.
I also created a PR for the EVA-CLIP model https://github.com/mlfoundations/open_clip/pull/265.
In https://github.com/mlfoundations/open_clip/pull/264, I add deepspped and zero-stage-1 optim. We can train 41k bsz with only 256 GPUs(A100-40G ) with grad checkpoint & deepspeed fp16 & zero-stage-1 .

Quan-Sun:
p.s. bsz can achieve 57k when using grad checkpoint & deepspeed fp16 & zero-stage-1 & local loss

rwightman:
@Quan-Sun @gabrielilharco I think the changes are reasonable, I use layer decay extensively in timm fine-tuning these days, and I feel it'd be useful here, especially when initializing one or both of the towers w/ pretrained weights.

Also, I was thinking of pushing optimizer creation into a factor method as well since I wanted to try some other optimizers at some point and that'd make it a bit cleaner.

Question though, why are the assigners in main? they're just created and passed to the optim factory, wouldn't it be cleaner to create them in the factory since they're not needed in main?

src/training/main.py
```diff
@@ -133,6 +133,22 @@ def main(args):
     )
     random_seed(args.seed, args.rank)
 
+    model_without_ddp = model
+    visual_ld = args.visual_ld if args.visual_ld else args.ld
+    text_ld = args.text_ld if args.text_ld else args.ld
+    
+    if visual_ld < 1.0:
+        visual_num_layers = model_without_ddp.visual.get_num_layers()
+        assigner_visual = LayerDecayValueAssigner(list(visual_ld ** (visual_num_layers + 1 - i) for i in range(visual_num_layers + 2)))
```
gabrielilharco, line 142: consider using a more descriptive variable name here, explaining what is being assigned (e.g. `lr_assigner_visual`). Also is there any reason why the exponent logic is not inside the Assigner class? I.e. the constructor could take in the layer decay and number of layers, and compute the values accordingly

src/training/main.py
```diff
@@ -133,6 +133,22 @@ def main(args):
...
+    else:
+        assigner_visual = None
+
+    if text_ld < 1.0:
```
gabrielilharco, line 146: should this be != 1.0?

src/training/optim.py
```diff
@@ -0,0 +1,133 @@
...
+        "visual.pos_embed", 
+        "visual.positional_embedding", 
+        "visual.patch_embed",
+        "visual.conv1", # name of patch embed in VisionTransformer
+        "text.pos_embed", 
+        "text.positional_embedding",
+        "text.token_embedding",
+        "text.transformer.embeddings.word_embeddings",
+        "text.transformer.embeddings.position_embeddings",
+        "text.transformer.embeddings.token_type_embeddings"
```
gabrielilharco, line 16: will this work with all architectures we support? E.g. text models from HF

src/training/optim.py
```diff
@@ -0,0 +1,133 @@
...
+
+    if len(parameters) == 0:
+        parameters = model.parameters()
+    return parameters
+
+def create_optimizer(args, model, assigner_visual, assigner_text):
+    parameters = get_all_parameters(args, model, assigner_visual, assigner_text)
+
+    optimizer_args = dict(
+            betas=(args.beta1, args.beta2),
```
gabrielilharco, line 128: nit: extra tab?

gabrielilharco:
Thanks for the context @Quan-Sun @rwightman. I agree with @rwightman re. the assigners, and also left some other minor comments there. I'll test it from my side after the changes. Thanks!

src/training/main.py
gabrielilharco: same for `visual_ld` above

src/training/optim.py
rwightman: @gabrielilharco no that's something else that needs to be addressed , as implemented it only works for built-in vit + text transformer, needs to at least detect if it will work and warn that it cannot be used for other models...

It'd be very useful for pretrained timm and HF text models, timm has functions that can calculate the layer decay but needs to be called if a timm model is used (that can be a diff PR), and not sure if HF has any built-in support to calculate layer-decay (discriminative LR) in a general way...

rwightman:
FWIW, timm's LD impl is here https://github.com/rwightman/pytorch-image-models/blob/e98c93264cde1657b188f974dc928b9d73303b18/timm/optim/optim_factory.py#L92-L153 ... all models have a fn that returns the group metadata, and the grouper fn can be used, so that would be basis for apply LD for vision tower if timm tower is used

src/training/main.py
rwightman: +1 on that, plus getting this out of main

Quan-Sun:
@gabrielilharco @rwightman Thanks for your comments. I will work on these changes ASAP.

gabrielilharco:
Thanks for the update @Quan-Sun. IIUC this would still only work for built-in vit + text transformers, is this right? As Ross pointed out, we should at least detect if this is not the case and warn users for models that are not supported

Quan-Sun:
Hi @gabrielilharco. You are right. get_num_layer_for_transformer(...) is not flexible. It should warn users if the models are not supported. Do you think we can have a white list here? For example, white_list = ["visual.blocks", "visual.transformer.resblocks", "text.transformer.resblocks", "text.transformer.encoder.layer"], then detecting if the model_param_name is in this white_list.

gabrielilharco:
Yes, that should work. If we're being very conservative we could also whitelist specific `args.model` values --- the downside is we'd need to update the list whenever we add new models

gabrielilharco:
Thanks for the update @Quan-Sun! Could you check if “Allow edits from maintainers.” is checked on your side? I want to do some small changes before merging

Quan-Sun:
Hi @gabrielilharco. have checked "Allow edits from maintainers." on my side. Please let me know if anything was missed.

PR #272, opened 2022-12-02
add patch dropout, as it has been proven out in new Kaiming He paper …

lucidrains:
…in the CLIP setting to save a ton of compute and improve end results

https://arxiv.org/abs/2212.00794


lucidrains:
Screenshot from 2022-12-02 09-30-47

figure says it all

src/open_clip/model.py
```diff
@@ -29,6 +29,7 @@ class CLIPVisionCfg:
     patch_size: int = 16
     image_size: Union[Tuple[int, int], int] = 224
     ls_init_value: Optional[float] = None  # layer scale initial value
+    patch_dropout: float = 0.  # what percentage of patches to dropout during training (not inference) - 0.5 to 0.75 recommended in the paper for optimal results
```
rom1504, line 32: this needs to be exposed through params.py

src/open_clip/transformer.py
```diff
@@ -255,6 +295,8 @@ def __init__(
         scale = width ** -0.5
         self.class_embedding = nn.Parameter(scale * torch.randn(width))
         self.positional_embedding = nn.Parameter(scale * torch.randn(self.grid_size[0] * self.grid_size[1] + 1, width))
+        self.patch_dropout = PatchDropout(patch_dropout)
```
rom1504, line 298: let's add a comment here saying that if patch_dropout is 0 then this does nothing

src/open_clip/transformer.py
```diff
@@ -336,6 +378,8 @@ def forward(self, x: torch.Tensor):
             [self.class_embedding.to(x.dtype) + torch.zeros(x.shape[0], 1, x.shape[-1], dtype=x.dtype, device=x.device),
              x], dim=1)  # shape = [*, grid ** 2 + 1, width]
         x = x + self.positional_embedding.to(x.dtype)
+
+        x = self.patch_dropout(x)
```
rom1504, line 382: here too

src/open_clip/model.py
rom1504: ah no I see, configured that way it would need to be in one of the model config json files

lucidrains:
comments added!

src/open_clip/model.py
lucidrains: it should default to 0. even if `patch_dropout` is omitted from all the json files?

rom1504: yes that's right

rom1504: I'm trying to think how we want to use that. Duplicating all the config files to be able to use it seems a little weird
maybe it should be a param in params.py and then we have a `model.patch_dropout = params.patch_dropout` somewhere

but let's not overthink it, I'll try it in the current state first

lucidrains: Yup, agreed! Let's see how effective it is in the Open clip setting

If it becomes a necessity, then we can add it to all the configs with the optimal dropout value

rom1504:
so I just tried it and I'm not seeing any speed up on B/32

rom1504:
maybe model is too small and I'm reaching some overhead, trying with L/14

rom1504:
no speed up with L/14 either (about 50 image/s/gpu on a100 bs 176 per gpu)

looking at the code, I'm wondering if we need some stronger masking to get some of the performance benefits ?

rom1504:
after asking the author, increased batch size, seeing +50% speed up at batch size 180k instead of 90k
will try to disable grad checkpointing and go back to 90k

lucidrains:
thanks for testing it out so soon and sharing your findings!

lucidrains:
my bad, i didn't account for an extra layer of indirection https://github.com/mlfoundations/open_clip/pull/272/files#diff-971a0ae4a19b0a452b4bf010c019f0adf0aa6adcef93a6d54e7dea409c8be7e7R109 🙏

rom1504:
Thanks for the fix!
It's ok, my run will finish in some 6h. That'll give us a baseline

lucidrains:
how do the updated experiments look?

rom1504:
Now there's indeed a significant speed up for large enough models (to go beyond the constant latency overhead) and large batches (to benefit from lower use thanks to patch drop out)

I'm running a B/32 (small one) at 1:1 settings with our other experiments to check results on metrics.
For now seems the loss is higher , which seems to be expected. Metrics looks ok

Will report results when I have them

(Will also do the fine tuning at no dropout as advised by paper)

lucidrains:
tremendous! :pray:  thank you for validating so quickly!

rom1504:
Oh yeah and maybe add a small section in readme

lucidrains:
> Oh yeah and maybe add a small section in readme

which section were you thinking?

lucidrains:
@rom1504 i made a guess to what you wanted https://github.com/mlfoundations/open_clip/pull/272/commits/6506fcb0a7d71a940ee4b10d2ab0e84b76fec747

lucidrains:
@ronghanghu @haooooooqi thank you for your great paper! it is so rare when something actually works :smile:

rom1504:
thanks, let's merge!

PR #273, opened 2022-12-04
whether to just average pool the output embeddings post transformer i…

lucidrains:
…n vision transformer, cls token is unnecessary based on a lot of follow up vision transformer works as well as from Beyer himself

https://arxiv.org/abs/2205.01580


rwightman:
@lucidrains agree, this is good to have in there.

PR #277, opened 2022-12-06
fix: cast dtype in visiontransformer

OrangeSodahub:
When using fp16 to inference the vision transformer, it occurs the dtype wrong: the input dtype mismatch the weights dtype.


rom1504:
@mitchellnw any thoughts on this ?

rwightman:
@OrangeSodahub please provide more detail as to how you're using it when you run into this issue, thanks

OrangeSodahub:
@rwightman You see, https://github.com/mlfoundations/open_clip/blob/5d683617c940b2723e19b097a228cac2f42c927e/src/open_clip/model.py#L189-L192
in `encode_text` you cast inputs dtype to align with the model, but you miss the same operation in `encode_image`. It's obvious that if your model is using fp16 but your input tensor is fp32 it will get errors right?

rwightman:
I've worked with a lot of vision models in PyTorch, and I can't recall seeing one that casts the input. What you say would work around this 'issue' if using the builtin ViT model and one forgets to cast the input, but would not work if a timm image tower is being used, or a torchvision, etc etc.

https://github.com/mlfoundations/open_clip/blob/5d683617c940b2723e19b097a228cac2f42c927e/src/training/zero_shot.py#L42-L46

PR #286, opened 2022-12-12
Lint.

rom1504:
pylint, mypy and black

Following https://github.com/rom1504/python-template


rom1504:
this will create merge conflicts in all PRs
so maybe we don't merge

rwightman:
@rom1504 for black, adding `--skip-string-normalization` is a little less opinionated and reduces diff quite a bit... 

But yeah, should probably focus on the major PR and some refactoring / design to make integrating them easier before activating this...

rom1504:
will redo later

PR #298, opened 2022-12-15
Update requirements.txt

AugustKarlstedt:
transformers seems to be required to use open_clip, not just for training. see https://github.com/mlfoundations/open_clip/blob/60534f4774b11a2532536f3f264a666f11fbb6b6/src/open_clip/hf_model.py#L102

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,4 +5,5 @@ ftfy
 tqdm
 huggingface_hub
 sentencepiece
+transformers
 protobuf==3.20.*


lopho:
This is an optional dependency and is only required to use models from huggingface.

AugustKarlstedt:
While it is optionally, it still should be included to be able to resolve dependencies e.g. in requirement lock files

rom1504:
I think it would make sense to simply clarify the error to explain to the
user they need to install transformer dependency if they use models coming
from huggingface.

PR #299, opened 2022-12-15
Matryoshka Representation Learning for Open CLIP

adityakusupati:
The PR hardcodes 4 granularities of dimensions to train CLIP on. 

While I hope the default hparams will work, it might be good to train it for slightly longer epochs.

### DIFF
--- a/src/open_clip/model.py
+++ b/src/open_clip/model.py
@@ -172,8 +172,9 @@ class CLIP(nn.Module):
         self.ln_final = text.ln_final
         self.text_projection = text.text_projection
         self.register_buffer('attn_mask', text.attn_mask, persistent=False)
-
-        self.logit_scale = nn.Parameter(torch.ones([]) * np.log(1 / 0.07))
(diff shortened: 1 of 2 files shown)


mitchellnw:
Hey Aditya thanks for the PR with MRL -- however if you want to make MRL an option it would be good to have a flag so that this PR does not affect default training and its not hard coded. Also, we'd want to do experiments to verify MRL works in this setting, which I can help you get set up on the UW compute -- we have a subset of LAION there.

adityakusupati:
Sure thing Mitchell. I was hard coding it so that Rom could use it to train a model. I can probably cleanup the code and make it a flag so that it does not break training. 

I don't think we will be merging this PR anyway. I can run things on LAION subset at UW with the generalized or hard coded variant.

mitchellnw:
sure can you convert to draft in the meantime?

adityakusupati:
Done.

PR #301, opened 2022-12-18
autoresume if experiment exists

mitchellnw:
If there is an experiment which already exists with the same `--name`, auto-resume from it instead of exiting.

### DIFF
--- a/src/training/main.py
+++ b/src/training/main.py
@@ -75,10 +75,10 @@ def main(args):
     if os.path.exists(args.log_path):
         if not args.save_most_recent:
             print(
-                "Error. Experimetn exists and can only autoresume if saving most recent."
+                "Error. Experiment exists and can only autoresume if saving most recent."
             )
             return -1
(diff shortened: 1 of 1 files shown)


rwightman:
@mitchellnw heh, I was actually thinking about this an hour ago...

* saving latest as done now is a bit error prone / slightly wasteful, it's done after the numbered checkpoint as a separate save, so possible it will fail with out of space when other succeed, should write once and hard link the other. It's also possible to corrupt the previous latest in a failure. So should write to numbered first and link to latest on success only. 
* But, after the above, don't actually need latest, we could just list and sort
* if we use the argument `--resume latest` it could 1. bypass the exists check as you've added here, do a list and sort to find the latest checkpoint, and resume from that 
*

mitchellnw:
Yea totally agree, and while I'll likely keep using this for my existing run I like your implementation better for the repo going forward so I'll close this. Thanks!

PR #303, opened 2022-12-19
PoC for a resume that searches for last checkpoint if `--resume latest` arg set

rwightman:
Re #301 

I haven't tested this yet, but will next time I get pre-empted

### DIFF
--- a/src/training/distributed.py
+++ b/src/training/distributed.py
@@ -1,6 +1,7 @@
 import os
 
 import torch
+import torch.distributed as dist
 
 try:
     import horovod.torch as hvd
@@ -111,3 +112,26 @@ def init_distributed_device(args):
     args.device = device
     device = torch.device(device)
     return device
+
+
+def broadcast_object(args, obj, src=0):
+    # broadcast a pickle-able python object from rank-0 to all ranks
+    if args.horovod:
+        return hvd.broadcast_object(obj, root_rank=src)
+    else:
+        if args.rank == src:
+            objects = [obj]
+        else:
+            objects = [None]
+        dist.broadcast_object_list(objects, src=src)
+        return objects[0]
+
+
+def all_gather_object(args, obj, dst=0):
+    # gather a pickle-able python object across all ranks
+    if args.horovod:
+        return hvd.allgather_object(obj)
+    else:
+        objects = [None for _ in range(args.world_size)]
+        dist.all_gather_object(objects, obj)
+        return objects
--- a/src/training/main.py
+++ b/src/training/main.py
@@ -1,5 +1,7 @@
+import glob
 import logging
 import os
+import re
 import sys
 import random
 from datetime import datetime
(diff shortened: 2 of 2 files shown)


rom1504:
where is it searching the last checkpoints ? what if 2 runs are running ?

rwightman:
@rom1504 args.checkpoint_path is constrained to the current name by default  `args.checkpoint_path = os.path.join(args.logs, args.name, "checkpoints")`

but the get_latest_checkpoint fn could be used to search across multiple folders if you had another scope that included multiple related runs (ie /myruns/run, /myruns/run_1, /myruns_run_2)... but called from main it's limited

rom1504:
this should be documented in the readme

gabrielilharco:
@rwightman I'm running into some errors when running this multi-node. The error happens when trying to call `broadcast_object` on `resume_from` here: https://github.com/mlfoundations/open_clip/pull/303/files#diff-8cac5527ae65d91d536016bb558349a70695c2e856a3e0526b21df7c69f9b8b2R143.

I changed the code so it only calls `broadcast_object` if `resume_from is None`, and it worked again. Should we do this?

Here's the error I get without the changes:

```
2023-01-11,01:17:15 | INFO | No latest resume checkpoint found in /mypath/to/my/checkpoints.                     
Traceback (most recent call last):                                                                                                                                                                         
  File "/fsx/home-gamaga/tng/tools/train.py", line 239, in <module>                                                                                                                                        
    success = main(main_args)                                                                                                                                                                              
  File "/fsx/home-gamaga/faceblur/open_clip/src/training/main.py", line 143, in main                                                                                                                       
    resume_from = broadcast_object(args, resume_from)                                                                                                                                                      
  File "/fsx/home-gamaga/faceblur/open_clip/src/training/distributed.py", line 126, in broadcast_object                                                                                                    
    dist.broadcast_object_list(objects, src=src)                                                                                                                                                           
  File "/fsx/home-gamaga/miniconda3/envs/tng/lib/python3.10/site-packages/torch/distributed/distributed_c10d.py", line 2081, in broadcast_object_list                                                      
    object_sizes_tensor = torch.empty(len(object_list), dtype=torch.long, device=current_device)                                                                                                           
RuntimeError: CUDA error: uncorrectable ECC error encountered                                                                                                                                              
CUDA kernel errors might be asynchronously reported at some other API call,so the stacktrace below might be incorrect.                                                                                     
For debugging consider passing CUDA_LAUNCH_BLOCKING=1. 
```

rwightman:
@gabrielilharco that looks li ke a hardware/cluster error not a code error, you can't add the None logic, all but the master are supposed to be None and ALL nodes need to make the same distributed call

rwightman:
it's the first distributed call now so if there are any issues w/ the interconnect it'll break there first instead of the first step

rwightman:
you can try commenting it out completely and doing a manual resume to see if resume + training works w/ the same node allocation and no broadcast_object, possible I got use of the call wrong but I've used it quite a bit now without issue from 16-1024 GPUs.

gabrielilharco:
ah ok, that's good to know, thanks for the info! Let me try some more runs to see if the issue persists. I wasn't getting this error with 2 nodes, it started showing up with 4, but great to hear you've used this with a large number of nodes.

PR #307, opened 2022-12-20
Add coca trained

rom1504:
trained branch

PR #308, opened 2022-12-20
Add coca trained (#307)

rom1504:
Proposal to merge coca into main.

This will be merged only once we are happy with results.

Reminder of what is still needed, as mentioned in https://github.com/mlfoundations/open_clip/pull/256 :+1: 

Priority:
* [x] implement and check captioning works  https://github.com/lucidrains/x-transformers/blob/main/x_transformers/autoregressive_wrapper.py
* [x] Make Coca use the visual encoder and text encoder of openclip fully
* [x] Revert changes to Attention / CustomResidualAttentionBlock, we should avoid using them with the CoCa model

Also important:
* [x] make sure normal clip is still working the same (#198 can help but otherwise just doing runs)
* [x] Move all models to output dicts instead of tuples, makes it more flexible pass output through optional loss / filters w/o brittle indexing and tuple len checks that are bound to fail with one more addition. Could make the dict output optional at construction time to prevent possible breaks for other downstream users.
* [x] report the captioning loss and contrastive loss independently in wandb

@gpucce is the main contributor on this
@iejMac is helping

I will review along with @rwightman 

@AwePhD and @Soonhwan-Kwon have proposed their help

we can train some models to validate things work

This is a first model https://huggingface.co/laion/CoCa-ViT-B-32-laion2B-s13B-b90k/blob/main/epoch_95.pt that gets same performance as clip (a bit better), see https://wandb.ai/rom1504/open-clip/reports/CoCa-ViT-B-32-v1-vs-clip--VmlldzozMTgzNTEy : 63.79% for coca vs 63.57% for a clip with same settings

For further improvements, please open PRs targeting the `coca` branch

This PR will be merged to main only when everything is good


rom1504:
The thing that would be most important to work on right now is captioning.
Using https://github.com/lucidrains/x-transformers/blob/main/x_transformers/autoregressive_wrapper.py should make it possible while using this model https://huggingface.co/laion/CoCa-ViT-B-32-laion2B-s13B-b90k/blob/main/epoch_95.pt 

we did some preliminary testing without proper AR sampling and it seems to do something which is good.

Now if this gets implemented, the basic are down and we can focus on the cleaning up tasks outlined above

rwightman:
Also pinging @gmittal and @apsdehal from FLAVA (#218), it looks like this will be a big merge prior to FLAVA being ready. So, some decisions here will impact that PR. One of those being a move to dict outputs from the model (I believe FLAVA has already done that).

We dict outputs we should  be able to have a fairly clean custom loss + logging mechanism that doesn't polute the main script that works for both.

gpucce:
@rom1504 `coca_base` is too large but I am not sure if `coca_ViT-B-32` is too big for github, if it is not it could be removed in `tests/test_inference.py` I can do it in a PR, thought I'd ask before adding noise

https://github.com/mlfoundations/open_clip/blob/50bc5991a82f601e14c155173e7c1ca536cd6af8/tests/test_inference.py#L13-L24

Soonhwan-Kwon:
@gpucce @rom1504  I'm middle of writing code for beam search and have a question that https://huggingface.co/laion/CoCa-ViT-B-32-laion2B-s13B-b90k/blob/main/epoch_95.pt is now outdated or not.

gpucce:
> @gpucce @rom1504 I'm middle of writing code for beam search and have a question that https://huggingface.co/laion/CoCa-ViT-B-32-laion2B-s13B-b90k/blob/main/epoch_95.pt is now outdated or not.

@Soonhwan-Kwon it is still the only one there is, and it should work almost as well as when it was trained, the only change introduced since it was trained is attention masking of padded tokens, however it only has a small effect on performance

gpucce:
@rom1504 I think adding https://github.com/sks3i/pycocoevalcap to the dependencies would make adding cider and other captioning metrics very easy, is it fine to add it? otherwise I can add similar code to a `eval_generation_utils.py` file

gpucce:
Initial super naive generation evaluation on coco:
|    |   Bleu_1 |   Bleu_2 |    Bleu_3 |    Bleu_4 |   METEOR |   ROUGE_L |    CIDEr |     SPICE |
|---:|---------:|---------:|----------:|----------:|---------:|----------:|---------:|----------:|
|  0 |  0.41722 | 0.181913 | 0.0506256 | 0.0130734 | 0.125147 |  0.285138 | 0.256736 | 0.0790488 |

Soonhwan-Kwon:
beam_search(num_beams=6) evaluation on coco:

| | Blue_1 | Blue_2 | Blue_3 | Blue_4 | METEOR | ROUGE_L | CIDEr | SPICE |
| ---- |---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- |
| 0 |  0.213 | 0.121 | 0.069 | 0.039 | 0.094 | 0.225 | 0.202 | (will be updated) |

src/open_clip/transformer.py
```diff
@@ -185,12 +213,33 @@ def __init__(
         ]))
         self.ls_2 = LayerScale(d_model, ls_init_value) if ls_init_value is not None else nn.Identity()
 
-    def attention(self, x: torch.Tensor, attn_mask: Optional[torch.Tensor] = None):
-        attn_mask = attn_mask.to(x.dtype) if attn_mask is not None else None
-        return self.attn(x, x, x, need_weights=False, attn_mask=attn_mask)[0]
+    def attention(
+        self,
```
rom1504, line 217: looks like this is not changing anything, let's revert it (same below)

agreed @gpucce @iejMac ?

rom1504:
looks good now

I think we need this now:
* [ ] check it doesn't break clip training
* [ ] document how to do coca training in readme
* [ ] train one small coca and release it
* [ ] fix sampling https://github.com/mlfoundations/open_clip/pull/320
* [ ] optionally: check what it would take to support other decoders eg some pretrained GPT from HF

rom1504:
would appreciate a second look @rwightman

rom1504:
This repo is moving fast these days, so we should try to either get this merged or get the most of the refactoring merged in the next week.
Otherwise there will be merge conflicts everywhere and it'll be annoying

rwightman:
@rom1504 this is still most significant change so I don't see the rush but will start picking through, conflicts are small relative to the commitment to maintain and continue getting this to work well. Is everyone involved in this definitely going to continue dev and get it to a useable/worthwhie point after merge? if it's merged and left with current result it'd be a net drag.

Without having dug in fully yet, my main concern is that this might break use of models in other train scripts? We should ensure that any changes to core modelling interface (ie dict outputs, tuple changes etc) are opt-in so that if there are notebooks, scripts out there just using 'open_clip' (modelling) they won't all break ...

rom1504:
> Without having dug in fully yet, my main concern is that this might break use of models in other train scripts? We should ensure that any changes to core modelling interface (ie dict outputs, tuple changes etc) are opt-in so that if there are notebooks, scripts out there just using 'open_clip' (modelling) they won't all break ...

We should try to add any method we consider as supported/exposed into the tests to make sure they keep working. That's the only reasonable path.
That and making it clear what is exposed and what is not

rom1504:
Regarding what other things it's going to conflict with, most likely text text clip, flava and fdsp. Those are less ready than coca but better get things in a good state early.

My remark was more in the direction of: we should either figure out how to merge this early, or if we can't, then it's better to extract the non-coca specific parts into another PR and merge those first, to make things easier. And I was saying that more for contributors than reviewers.

gpucce:
@rom1504 @rwightman For info I will continue working on this until it is at a good point as far as it is needed (hoping my effort gets the whole thing somewhere) regardless of merge.

Besides this, I think there should be one run going, when it finishes, if clip evaluation is ok, I will try and fine tune on coco train dataset to replicate evaluation as it is done in the coca paper and see if we get somewhere closer to the original.

rwightman:
if there's a commitment from @gpucce @iejMac to keep pushing this to a desired performance level once it's merged (and not just let it dangle), we can make a push to try and merge in the comming week. Need to figure out the bwd compat concerns, and also take a closer look at the direction Flava is headed wrt to the dictionary outputs and loss fn to ensure we have compatible directions there...

gpucce:
checking the FLAVA PR I list here the main differences to choose the way to go:
- [ ] outputting dicts is an option in COCA while it is enforced in FLAVA both for losses and models;
- [ ] COCA defines an extra transformer in `transformer.py` FLAVA defines it in `flava_model.py`;
- [ ] FLAVA defines a `Batch` class which should be compatible with `COCA`

iejMac:
@rwightman I'm definitely interested in pushing performance. As to merge readiness I think the main question is how do we ensure normal functionality isn't broken. Other stuff is more or less handled:
- B/32 CoCa is currently training https://wandb.ai/iejmac/open-clip/reports/CoCa-ViT-B-32-v2--VmlldzozMjM0Nzg1
- Documenting in README won't be hard
- sampling PR seems to be almost ready? Maybe @Soonhwan-Kwon  can comment on the state of it?https://github.com/mlfoundations/open_clip/pull/341
- I think we can leave pretrained huggingface decoder compatibility for later

Soonhwan-Kwon:
> @rwightman I'm definitely interested in pushing performance. As to merge readiness I think the main question is how do we ensure normal functionality isn't broken. Other stuff is more or less handled:
> 
> * B/32 CoCa is currently training https://wandb.ai/iejmac/open-clip/reports/CoCa-ViT-B-32-v2--VmlldzozMjM0Nzg1
> * Documenting in README won't be hard
> * sampling PR seems to be almost ready? Maybe @Soonhwan-Kwon  can comment on the state of it?[add beam search,fix image_latent,image_emb calculated once #341](https://github.com/mlfoundations/open_clip/pull/341)
> * I think we can leave pretrained huggingface decoder compatibility for later

Adding beam search has been done, but w/ past_key_values to generate fast is in progress. Generation w/o past_key_values is little bit slower, but I think it is enough for getting benchmark outputs.

gpucce:
@iejMac @Soonhwan-Kwon @rom1504 @rwightman 

With 6 epoch of fine-tuning on mscoco, with the very first model we trained and with beam_search I get 

|         | 0                  | 1                  | 2                   | 3                 | 4                  | 5                   | 6                 | 7                   |
|:--------|:-------------------|:-------------------|:--------------------|:------------------|:-------------------|:--------------------|:------------------|:--------------------|
| index   | Bleu_1             | Bleu_2             | Bleu_3              | Bleu_4            | CIDEr              | METEOR              | ROUGE_L           | SPICE               |
| metrics | 0.6625615763546681 | 0.5142596473045911 | 0.39456296262989304 | 0.303324191412218 | 0.8187631205735131 | 0.24591518949678903 | 0.535950978552973 | 0.20529777849882702 |


Which is still a lot worse than the CoCa paper but a lot less terrible, if I want to make fine-tuning part of the repos should I add it here or in CLIP_benchmark?

rom1504:
Let's add it in clip benchmark

rwightman:
looking better! for fine-tune, how much extra code is it? I feel fine-tune falls closer to a train focused repo than benchmarking?

gpucce:
> looking better! for fine-tune, how much extra code is it? I feel fine-tune falls closer to a train focused repo than benchmarking?

I did it on open_clip creating a csv from the coco train split and this way it is the same as pretraining with different data and hyperparams. To add to clip_benchmark I am trying to see if I can run `training.main` to reuse most of the code. 

There is one benchmark that does extra training in clip_benchmark already, but it is different I think and the training code is reimplemented from zero.

rwightman:
Has the coca model been tested to work with torchscript (jit=True) in inference mode? Come to think of it has the main CLIP model been tested with the changes? Thinks like dict or tuple, tensor or tuple outputs have a tendency to be not so fun w/ torchscript. 

I thought we added jit = True to some of the tests but looks like maybe not. Whenever we break torchscript someone invariably files a bug as it's still used in many hosting/export scenarios.

rom1504:
https://github.com/mlfoundations/open_clip/blob/main/tests/test_inference.py#L42 could be checked by putting this to True

It's not done by default because
* On CPU (gh action), jit is very slow (many minutes)
* on GPU, jit has much slower initialization (1-2min) and no speed gain

gpucce:
@Soonhwan-Kwon I looked into adding beam_search (which is great as validation works a lot better!) and adding the `past_key_values` option seems to mean adding several changes, many in `transformer.py`, if you agree with this and you are not too far ahead implementing it locally, maybe you can remove the few changes you already made to work toward it and so the PR can be merged more easily?

This is just an idea and if you or anybody else here thinks otherwise I am absolutely ok with it.

Soonhwan-Kwon:
> @Soonhwan-Kwon I looked into adding beam_search (which is great as validation works a lot better!) and adding the `past_key_values` option seems to mean adding several changes, many in `transformer.py`, if you agree with this and you are not too far ahead implementing it locally, maybe you can remove the few changes you already made to work toward it and so the PR can be merged more easily?
> 
> This is just an idea and if you or anybody else here thinks otherwise I am absolutely ok with it.

Sure, I'll remove changes related to past_key_values.

Soonhwan-Kwon:
@gpucce removed changes related to past_key_values in https://github.com/mlfoundations/open_clip/pull/341

iejMac:
@gpucce @Soonhwan-Kwon new fully trained CoCa model is up on hf https://huggingface.co/iejMac/CoCa_tests feel free to evaluate it

rwightman:
> https://github.com/mlfoundations/open_clip/blob/main/tests/test_inference.py#L42 could be checked by putting this to True
> 
> It's not done by default because
> 
> * On CPU (gh action), jit is very slow (many minutes)
> * on GPU, jit has much slower initialization (1-2min) and no speed gain

Right, yeah it can be slow, but I jit test lots of models and the small ones shouldn't take that long (ie a B/32). Despite it not being a significant improvement for throughput, many people use it for torch serving, or for loading in C++ (LibTorch).

rom1504:
Sounds good, let's add that to tests then

gpucce:
@iejMac @rom1504 @rwightman @Soonhwan-Kwon 

new results with newest model 

| index   | Bleu_1             | Bleu_2             | Bleu_3            | Bleu_4              | CIDEr            | METEOR            | ROUGE_L           | SPICE               |
|:--------|:-------------------|:-------------------|:------------------|:--------------------|:-----------------|:------------------|:------------------|:--------------------|
| metrics | 0.7084 | 0.5413 | 0.4073 | 0.3041 | 1.0273 | 0.2695 | 0.5368 | 0.2048 |

Can probably be pushed further just by better cleaning/processing the generated strings only or with better generation strategy.

gpucce:
> Sounds good, let's add that to tests then

should I add in this PR?

rom1504:
https://wandb.ai/iejmac/open-clip/reports/CoCa-PR-sanity-check--VmlldzozMzg3OTcz sanity check of clip. looks ok

rom1504:
This is pretty good now.
We're thinking to merge soon.

Maybe integrating the pretrained model too

src/open_clip/coca_model.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def encode_text(self, text, normalize=True):
+        text = text[:, :-1] # make space for CLS token
+        text_latent, token_embs = self.text.encoder(text)
+        text_latent = F.normalize(text_latent, dim=-1) if normalize else text_latent
+        if getattr(self.text.encoder, "output_tokens", False):
+            return text_latent, token_embs
+        return text_latent
+    
+    def forward(self, image, text, output_dict=False):
```
rwightman, line 139: what's output_dict doing here?

src/open_clip/model.py
```diff
@@ -209,6 +228,12 @@ def encode_text(self, text, normalize: bool = False):
     def forward(self, image, text):
         image_features = self.encode_image(image, normalize=True)
         text_features = self.encode_text(text, normalize=True)
+        if self.output_dict is not None:
+            return {
```
rwightman, line 232: why is this None and not bool? bool attr should work fine with torchscript

Soonhwan-Kwon:
@rom1504 @rwightman I believe that tests was broken because coca has output_dict in forward and train.py has some forward with output_dict and otherwise some are not. I tried to fix thoes incosistency with clip in https://github.com/mlfoundations/open_clip/pull/341 with introducing output_dict in coca initalization, and also is_coca, postprocess_coca_output.

src/open_clip/coca_model.py
iejMac: shouldn't be there, here's the solve: https://github.com/mlfoundations/open_clip/pull/371

src/open_clip/model.py
iejMac: @gpucce tried changing that here: https://github.com/mlfoundations/open_clip/pull/373

src/open_clip/coca_model.py
```diff
@@ -0,0 +1,209 @@
...
+    mlp_ratio: int = 4
+    dim_head: int = 64
+    heads: int = 8
+    n_queries: int = 256
+    attn_pooler_heads: int = 8
+    latent_dim: int = 512
+
+
+class CoCaEncoderDecoder(nn.Module):
+    def __init__(self, encoder, decoder) -> None:
```
rwightman, line 30: Is there are reason for this module to exist? vs a text_encoder and tex_decoder in CoCa model? having modules organized like this (two modules in a class with no forward that exercises) is atypical and breaks some assumptions for profilers, etc

rwightman: FYI, checkpoints can easily be remapped, so existing checkpoints is not a reason not to make changes right now

src/open_clip/coca_model.py
```diff
@@ -0,0 +1,209 @@
...
+
+        self.text, multimodal_cfg, vocab_size = _build_encoder_decoder_tower(
+            embed_dim, multimodal_cfg, text_cfg, quick_gelu, cast_dtype
+        )
+        self.visual = _build_vision_tower(
+            multimodal_cfg.latent_dim, vision_cfg, quick_gelu, cast_dtype
+        )
+
+        self.to_logits = nn.Sequential(
+            norm_layer(multimodal_cfg.width), nn.Linear(multimodal_cfg.width, vocab_size, bias=False)
```
rwightman, line 113: I'm still not a fan of these being in a nn.Sequential with no names, should either be separate attributes like `.decoder_norm`, `.decoder_logits`, or add names like 'norm', 'fc'  to the nn.Sequential, without that you'd break checkpoint compat if you say wanted to experiment and add a dropout btw norm and linear

src/open_clip/coca_model.py
```diff
@@ -0,0 +1,209 @@
...
+        self.decoder.set_grad_checkpointing(enable)
+
+
+def _build_encoder_decoder_tower(
+        embed_dim,
+        multimodal_cfg,
+        text_cfg,
+        quick_gelu: bool = False,
+        cast_dtype: Optional[torch.dtype] = None,
+):
```
rwightman, line 47: As per the Encoder/Decoder Module above, with those split, this can be split to have text(_encoder) + text_decoder

src/open_clip/coca_model.py
rwightman: More thoughts, to keep it a bit closer to the Clip model on the encoder side 

```
self.text = _build_text_encoder()
self.visual = _build_vision_tower()
self.decoder = _build_multimodal_decoder() # or self.multimodal_decoder or .text_decoder?
self.decoder_norm = 
self.decoder_logits =
```

rwightman:
Is the current coca_ViT-B-32 model expected to work with zero-shot imagenet eval? it's broken due to tuple output

rwightman:
@gpucce regarding some of the inline comments above, I had a chat with @iejMac and @rom1504 

@iejMac pointed out I might have been a bit misleading whan I said 'make it like the dual tower CLIP model', sorry about that, I meant make like that in terms of encapsulating related logic within each tower (vs original CLIP that had the text transformer parts spilling into the main model), so in this case I feel 3 towers would be appropriate

`model.text` - text encoder
`model.visual` - image encoder
`model.text_decoder` - multi-modal input, unimodal output decoder (we had some differing opinions here, in `model.multimodal`, `model,decoder`, `model.text_decoder`, the last one seemed the most favoured, do you have any different opinions / ideas?

With the above, the other highest priority thing is the to_logits. It looks like there is a bug there, the multimodal transformer ends with a LayerNorm, and to_logits begins with a layernorm, so that's two back to bcak layernorms where I think there should be just one. Are there any cases where there should be logic / pooling etc between those two LN? Also related, we should push this into the text_decoder module instead of having it in the base model (unless there was a definite reason to have it factored out like that?)

gpucce:
@rwightman I am doing it exactly this way.

The double layernorm is indeed a bug I thought I had removed the one in coca_model. If I understand that is the one you want removed?

I have a PR almost ready will open it in a moment and finish it while it is open, today was busier than I hoped, I am checking that the refactored model still works with mscoco (this does) and with imagenet this I am doing it currently.

gpucce:
> I was thinking we leave the final LN in the decoder transformer module and put the final projection from the coca model to_logits into the decoder as well, similar to the final proj in the other transformer towers. Does that make sense? Are there normal uses where you wouldn't want the final projection to be performed?

The only case I can think of is maybe someone might want to use the output of the decoder also as an embedding to feed to some other model/classifier but probably it is not really meaningful anyway. Maybe conceptually one might want to make it different from the project in the other models as the to_logits is supposed to map to very large vocabulary-size vectors while the others, if I understand, are always used to map to hidden states in contrastive loss.

Good news is that removing the layernorm also from the existing pretrained model is completely harmless at least with respect to zeroshot iamgenet checking with mscoco in a while, will share the refactored model too once the rest is done

rwightman:
If that's the case we should push that nn. Linear into the decoder module.
Can always add an extra arg to forward to output pre-logits if it's needed
in the future, but makes it cleaner otherwise to encapsulate that layer
with rest of decoder

gpucce:
> If that's the case we should push that nn. Linear into the decoder module. Can always add an extra arg to forward to output pre-logits if it's needed in the future, but makes it cleaner otherwise to encapsulate that layer with rest of decoder

Yeah I moved it already, pushing in a bit after checking benchmarks are both ok still

gpucce:
@rwightman

#378 should do all the things that you have suggested. I tested the refactored model using the pretrained one zeroshot imagenet stays the same as changes are only in the decoder part (checked anyway) and mscoco_generative is now as follows compared to best so far

| index   | Bleu_1             | Bleu_2             | Bleu_3            | Bleu_4              | CIDEr            | METEOR            | ROUGE_L           | SPICE               |
|:--------|:-------------------|:-------------------|:------------------|:--------------------|:-----------------|:------------------|:------------------|:--------------------|
| best | 0.708 | 0.541 | 0.407 | 0.304 | 1.027 | 0.27 | 0.537 | 0.205 |
| minus layernorm | 0.699 | 0.524 | 0.388 | 0.288 | 0.977 | 0.263 | 0.532 | 0.203 |

(My home connection is working very poorly for some reason so I will share the refactored `state_dict` tomorrow morning)

rwightman:
@gpucce thanks! looks great and merged to this PR. About to merge the jit tweaks before I do some more local testing, I think that jit approach is good for now

rom1504:
Are there any next steps here ?
Any reason not to press merge ?

rwightman:
@rom1504 I think it's pretty much ready, I suppose should go through one more check of the jit stuff since I haven't re-tested since last few additions

I was watching the L/14 run that was active to see if there is any reason to be concerned

gpucce:
> Are there any next steps here ?
> 
> Any reason not to press merge ?

Maybe finish generate_beamsearch? For the evaluation of the model I am using it. @Soonhwan-Kwon can I help somehow?

Soonhwan-Kwon:
@gpucce It's a pleasure, if you help me. I gave you invitation for collaborator in my repo.

rom1504:
up to date with main branch now

rwightman:
Looking at this now, running into some sort of None / dtype issue in MHA in JIT, not quite sure what's going on just yet...

gpucce:
> Looking at this now, running into some sort of None / dtype issue in MHA in JIT, not quite sure what's going on just yet...

Could be this https://github.com/pytorch/pytorch/issues/92073 maybe?

rwightman:
> > Looking at this now, running into some sort of None / dtype issue in MHA in JIT, not quite sure what's going on just yet...
> 
> Could be this [pytorch/pytorch#92073](https://github.com/pytorch/pytorch/issues/92073) maybe?

Yup, it's exactly that and it's been a known issue for over a year! https://github.com/pytorch/pytorch/issues/71470

FFS

src/open_clip/coca_model.py
```diff
@@ -0,0 +1,194 @@
...
+    )
+
+    return decoder
+
+
+class CoCa(nn.Module):
+    def __init__(
+            self,
+            embed_dim,
+            multimodal_cfg: MultimodalCfg,
```
rwightman, line 59: I just noticed this, embed_dim isn't used for CoCa as it's taken from multimodal_cfg.latent_dim ... a little bit weird to have the values in cfg, and the arg, and then not use it.. hmmm

rwightman: It doesn't look like the MultimodalTransformer tower uses the latent_dim itself, so should that just be the determined by the cfg['embed_dim'] like the other models and remove multimodal_cfg['latent_dim'] ?

rwightman: @gpucce

rwightman:
For JIT, given the only issue I see right now is a PyTorch bug (that should be easy to fix, and is CoCa specific) I don't see any issues there... hopefully it gets fixed soon in PT.

rwightman:
I think we are close, I have another bit of logic to visit in TextTransformer, that embed_dim above in CocaModel to resolve

src/training/train.py
```diff
@@ -94,17 +97,33 @@ def train_one_epoch(model, data, epoch, optimizer, scaler, scheduler, args, tb_w
 
         if args.accum_freq == 1:
             with autocast():
-                image_features, text_features, logit_scale = model(images, texts)
-                total_loss = loss(image_features, text_features, logit_scale)
+                model_out = model(images, texts)
+                # for clip if it does not output_dict
+                module = model.module if type(model) == DistributedDataParallel else model
+                if is_clip(module) and not module.output_dict:
+                    model_out = postprocess_clip_output(model_out)
```
rwightman, line 104: that last thing of note in my list, I don't feel there is any need to supoort clip not outping a dict
* allowing both output modes in the model was that so users in other training codebases, or in existing colab notebooks, etc wouldn't have full break in their code when they update open_clip
* our training code can always operate in outpu_dict mode, I don't see a need to support both
* if anyone is using open_clip train code with their own models, they can adapt it easily enough (they'll be more advanced users)

rom1504: makes sense, can you create an issue for it ?

src/open_clip/coca_model.py
rom1504: is it resolved ? if not can you create an issue for it ?

rom1504:
thanks for all the work on coding this @gpucce !

thanks @iejMac for running the experiments !

and thanks @rwightman for reviews!

rom1504:
Please create dedicated issues for follow ups

rom1504:
currently known follow ups listed at https://github.com/mlfoundations/open_clip/issues/390

PR #312, opened 2022-12-21
Add cls mask

gpucce:
this is supposed to add a mask to the attention preventing the CLS token at the end of the sequence from attendin to padded tokens.


rom1504:
that seems to make sense

we noticed perf changes a bit post this change though, so I figure this changes the beavior of the model and will need retraining

I guess we can merge it though ?

PR #313, opened 2022-12-21
simplify encode_image

gpucce:
This is supposed to make the `encode_image` of CoCa simpler without breaking things with respect to the currently trained model 

https://huggingface.co/laion/CoCa-ViT-B-32-laion2B-s13B-b90k/blob/main/epoch_95.pt 

it might not be the best way but I thought I would open the PR anyway to see which way it should be done.


rom1504:
makes sense!

rom1504:
what about the same for the text encoder ?

PR #319, opened 2022-12-23
Utility for sycning with s3 and loading checkpoints from s3

mitchellnw:
This PR introduces two additional arguments, which are `--sync-s3` and `--sync-s3-frequency`. Recommended use is to do `--sync-s3 s3://<path-to-bucket>` and `--logs /scratch/logs` which is hopefully local ssd.

Then, as you run, you should see logs at `/scratch/logs/<name>` and `s3://<path-to-bucket>/<name>`. So you don't have to use the local file system.

This PR also supports loading from `s3://<path-to-checkpoint>` -- it's a bit slow but not too bad.

### DIFF
--- a/src/training/main.py
+++ b/src/training/main.py
@@ -33,7 +33,7 @@ from training.logger import setup_logging
 from training.params import parse_args
 from training.scheduler import cosine_lr
 from training.train import train_one_epoch, evaluate
-from training.file_utils import pt_load, check_exists, start_sync_process, sync_s3
+from training.file_utils import pt_load, check_exists, start_sync_process, remote_sync
 
 
 LATEST_CHECKPOINT_NAME = "epoch_latest.pt"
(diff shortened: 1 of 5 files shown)


rom1504:
What about using fsspec for the same thing to avoid being locked on S3 ?

rom1504:
Can you add a comment it makes sense to use --sync-s3 with a local --logs in the readme ?

mitchellnw:
Thanks for the comments, updated! The default is syncing via `aws s3 sync`.

rom1504:
@mitchellnw could you resolve the merge conflict ?

do you think we should merge ?

mitchellnw:
ah, yea I should make this work with the new auto-resume which is where the conflict is coming from (https://github.com/mlfoundations/open_clip/pull/303). then yes I think good to merge after that

mitchellnw:
merge conflict fixed but need to add support for the resume = 'latest' feature

mitchellnw:
Ok, should be good to go.

src/training/main.py
```diff
@@ -49,9 +51,16 @@ def natural_key(string_):
     return [int(s) if s.isdigit() else s for s in re.split(r'(\d+)', string_.lower())]
 
 
-def get_latest_checkpoint(path: str):
+def get_latest_checkpoint(path: str, remote : bool):
     # as writen, this glob recurses, so can pick up checkpoints across multiple sub-folders
-    checkpoints = glob.glob(path + '**/*.pt', recursive=True)
+    if remote:
+        result = subprocess.run(["aws", "s3", "ls", path + "/"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
rom1504, line 57: would be much better to use fsspec to do that kind of listing

rom1504:
looks ok, let's go

PR #321, opened 2022-12-24
use `TextEncoder` in coca `encode_image`

gpucce:
This PR adds the possibility to use a CLS token as a model parameter in TextEncoder to simplify the logic of `CoCa`

### DIFF
--- a/src/open_clip/coca_model.py
+++ b/src/open_clip/coca_model.py
@@ -84,8 +84,8 @@ class CoCa(nn.Module):
 
         text = _build_input_dependent_text_tower(embed_dim, text_cfg, quick_gelu, cast_dtype, multimodal=False)
         self.text = text
-        self.context_length = self.text.positional_embedding.shape[0]
-        self.heads = text_cfg["heads"]
+        if "hf_model_name" not in text_cfg or text_cfg["hf_model_name"] is None:
+            self.text_projection = nn.Parameter(torch.randn(embed_dim, embed_dim))
         
         self.visual = _build_vision_tower(
             embed_dim, vision_cfg, quick_gelu, cast_dtype
@@ -100,13 +100,18 @@ class CoCa(nn.Module):
         )
 
         self.img_attn_pool_norm = norm_layer(embed_dim)
+        vocab_size = (
+            text_cfg["vocab_size"] 
+            if "vocab_size" in text_cfg
+            else self.text.config.vocab_size # for hf models
+        )
 
         self.to_logits = nn.Sequential(
-            norm_layer(embed_dim), nn.Linear(embed_dim, self.text.vocab_size, bias=False)
+            norm_layer(embed_dim), nn.Linear(embed_dim, vocab_size, bias=False)
         )
 
         # tie embedding weights and projection
-        self.to_logits[-1].weight = self.text.token_embedding.weight
+        # self.to_logits[-1].weight = self.text.token_embedding.weight
 
         self.logit_scale = nn.Parameter(torch.ones([]) * np.log(1 / 0.07))
         self.pad_id = 0
@@ -139,10 +144,11 @@ class CoCa(nn.Module):
 
     def encode_text(self, text, normalize=True, return_tokens=False):
         text = text[:, :-1] # make space for CLS token
-        cls_emb, token_emb = self.text(text, output_tokens=True)
+        text_latent, token_emb = self.text(text, output_tokens=True)
 
-        if hasattr(self.text, "text_projection") and self.text.text_projection is not None:
-            text_latent = cls_emb @ self.text.text_projection
+        # not HF model
+        if hasattr(self, "text_projection") and self.text_projection is not None:
+            text_latent = text_latent @ self.text_projection
 
         text_latent = F.normalize(text_latent, dim=-1) if normalize else text_latent
 
@@ -172,7 +178,7 @@ class CoCa(nn.Module):
         image,
         text,
         seq_len,
-        max_seq_len=None,
+        max_seq_len=77,
         mask_prob = 0.0,
         temperature = 1.,
         filter_logits_fn = top_k,
@@ -182,9 +188,6 @@ class CoCa(nn.Module):
         ):
 
         assert mask_prob < 1, "mask_prob must be smaller than 1."
-
-        if max_seq_len is None:
-            max_seq_len = self.context_length
 
         was_training = self.training
         num_dims = len(text.shape)
--- a/src/open_clip/coca_model.py
+++ b/src/open_clip/coca_model.py
@@ -22,7 +22,8 @@ class MultimodalCfg(CLIPTextCfg):
     dim_head: int = 64
     heads: int = 8
     n_queries: int = 256
-    dim_latents: int = None
+    attn_pooler_heads: int = 8
+    latent_dim: int = 512
 
 
 def _build_input_dependent_text_tower(
@@ -82,34 +83,32 @@ class CoCa(nn.Module):
             else LayerNorm
         )
 
-        text = _build_input_dependent_text_tower(embed_dim, text_cfg, quick_gelu, cast_dtype, multimodal=False)
-        self.text = text
-        if "hf_model_name" not in text_cfg or text_cfg["hf_model_name"] is None:
-            self.text_projection = nn.Parameter(torch.randn(embed_dim, embed_dim))
-        
-        self.visual = _build_vision_tower(
-            embed_dim, vision_cfg, quick_gelu, cast_dtype
-        )
-
         self.multimodal_decoder, multimodal_cfg = _build_input_dependent_text_tower(
             embed_dim, multimodal_cfg, quick_gelu, cast_dtype
         )
 
+        text = _build_input_dependent_text_tower(multimodal_cfg.width, text_cfg, quick_gelu, cast_dtype, multimodal=False)
+        self.text = text
+        self.visual = _build_vision_tower(
+            multimodal_cfg.width, vision_cfg, quick_gelu, cast_dtype
+        )
         self.img_attn_pool = AttentionalPooler(
-            multimodal_cfg.width, multimodal_cfg.heads, n_queries=n_queries + 1 # extra query for contrastive_loss
+            multimodal_cfg.width, multimodal_cfg.attn_pooler_heads, n_queries=n_queries + 1 # extra query for contrastive_loss
         )
 
-        self.img_attn_pool_norm = norm_layer(embed_dim)
+        self.img_attn_pool_norm = norm_layer(multimodal_cfg.width)
         vocab_size = (
-            text_cfg["vocab_size"] 
-            if "vocab_size" in text_cfg
-            else self.text.config.vocab_size # for hf models
+            self.text.config.vocab_size
+            if "hf_model_name" in text_cfg and text_cfg["hf_model_name"] is not None
+            else multimodal_cfg.vocab_size # for hf models
         )
 
         self.to_logits = nn.Sequential(
-            norm_layer(embed_dim), nn.Linear(embed_dim, vocab_size, bias=False)
+            norm_layer(multimodal_cfg.width), nn.Linear(multimodal_cfg.width, vocab_size, bias=False)
         )
 
+        self.to_img_latent = nn.Linear(multimodal_cfg.width, multimodal_cfg.latent_dim, bias=False)
+        self.to_txt_latent = nn.Linear(multimodal_cfg.width, multimodal_cfg.latent_dim, bias=False)
         # tie embedding weights and projection
         # self.to_logits[-1].weight = self.text.token_embedding.weight
 
@@ -125,7 +124,7 @@ class CoCa(nn.Module):
     def encode_image(self, images, normalize=True, return_tokens=False):
         x = self.visual(images, output_tokens=True)
 
-        if hasattr(self.visual, "ln_post"):
+        if hasattr(self.visual, "ln_post") and self.visual.ln_post is not None:
             x = self.visual.ln_post(x)
 
         if hasattr(self.visual, "proj") and self.visual.proj is not None:
@@ -134,7 +133,7 @@ class CoCa(nn.Module):
         x = self.img_attn_pool(x, x)
         x = self.img_attn_pool_norm(x)
 
-        image_latent = x[:, 0]
+        image_latent = self.to_img_latent(x[:, 0])
         image_latent = F.normalize(image_latent, dim=-1) if normalize else image_latent
 
         return (image_latent, x[:, 1:]) if return_tokens else image_latent
@@ -144,12 +143,9 @@ class CoCa(nn.Module):
 
     def encode_text(self, text, normalize=True, return_tokens=False):
         text = text[:, :-1] # make space for CLS token
-        text_latent, token_emb = self.text(text, output_tokens=True)
+        cls_emb, token_emb = self.text(text, output_tokens=True)
 
-        # not HF model
-        if hasattr(self, "text_projection") and self.text_projection is not None:
-            text_latent = text_latent @ self.text_projection
-
+        text_latent = self.to_txt_latent(cls_emb)
         text_latent = F.normalize(text_latent, dim=-1) if normalize else text_latent
 
         return (text_latent, token_emb) if return_tokens else text_latent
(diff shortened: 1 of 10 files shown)


gpucce:
@rom1504 please don't merge this yet

src/open_clip/coca_model.py
```diff
@@ -148,43 +150,10 @@ def encode_image(self, images, normalize=True, return_tokens=False):
...
-        
-        x = x.permute(1, 0, 2)  # NLD -> LND
-        x = self.transformer(x, attn_mask=attn_mask + cls_mask)
-        x = x.permute(1, 0, 2)  # LND -> NLD
-
-        x = x[torch.arange(x.shape[0]), :] @ self.text_projection
-
-        cls_emb = x[torch.arange(x.shape[0]), -1]
-        token_emb = x[torch.arange(x.shape[0]), :-1]
+        cls_emb, token_emb = self.text(text, output_tokens=True)
```
rom1504, line 155: I guess this misses a .encode_text

gpucce: The idea would be that `self.text` is a `TextTransformer` intialized with `embed_cls=True` and so the forward outputs the extra CLS embedding and this is used within `encode_text` so using `self.text` should be ok, If I understand what you mean

rom1504: Ok

src/open_clip/coca_model.py
```diff
@@ -76,56 +76,45 @@ def __init__(
...
-        self.token_embedding = text.token_embedding
-        self.positional_embedding = text.positional_embedding
-        self.ln_final = text.ln_final
-        self.text_projection = text.text_projection
-        self.register_buffer("attn_mask", text.attn_mask, persistent=False)
-        self.context_length = self.positional_embedding.shape[0] - 1
-
-        self.cls_token = nn.Parameter(torch.randn(embed_dim))
+        self.text = text
+        self.context_length = self.text.positional_embedding.shape[0]
```
rom1504, line 87: This is available in the config as well

src/open_clip/coca_model.py
```diff
@@ -76,56 +76,45 @@ def __init__(
...
         )
 
         self.img_attn_pool_norm = norm_layer(embed_dim)
 
-        self.dim_latents = multimodal_cfg.dim_latents if multimodal_cfg.dim_latents else multimodal_cfg.width
-        self.to_text_latent = nn.Linear(embed_dim, self.dim_latents, bias=False)
-
         self.to_logits = nn.Sequential(
-            norm_layer(embed_dim), nn.Linear(embed_dim, self.vocab_size, bias=False)
+            norm_layer(embed_dim), nn.Linear(embed_dim, self.text.vocab_size, bias=False)
```
rom1504, line 105: This is available in the config

Try not to rely on any self.text.x attributes as they might not be available for HF transformers

src/open_clip/coca_model.py
```diff
@@ -76,56 +76,45 @@ def __init__(
...
         )
 
         # tie embedding weights and projection
-        self.to_logits[-1].weight = self.token_embedding.weight
+        self.to_logits[-1].weight = self.text.token_embedding.weight
```
rom1504, line 109: Is this needed ?

gpucce: This is supposed to be how they do, however it should not be necessary, I removed it for the moment

gpucce:
@rom1504 this should be ready, essentially it addresses these points that were raised in the old PR and consequently adds possibility for text only encoder to be the Huggingface one, to make the multimodal one also available as HF is more difficult and I would do it in a different PR. As usual reviews, comments and anything else are more than welcome.

> would appreciate your review @rwightman if you think anything big need to be done

Made some code review comments, most important points:
* Make CoCa model dual tower (.visual, .text) from the start so builtin and HF text transformer models are the same, no bwd compat to be concerned about
* Revert changes to Attention / CustomResidualAttentionBlock, we should avoid using them with the CoCa model
* Move all models to output dicts instead of tuples, makes it more flexible pass output through optional loss / filters w/o brittle indexing and tuple len checks that are bound to fail with one more addition. Could make the dict output optional at construction time to prevent possible breaks for other downstream users.

and then test test test.

_Originally posted by @rwightman in https://github.com/mlfoundations/open_clip/issues/256#issuecomment-1356891072_

src/open_clip/coca_model.py
```diff
@@ -76,62 +77,54 @@ def __init__(
...
-            embed_dim, vision_cfg, quick_gelu, cast_dtype
-        )
-        self.heads = text_cfg["heads"]
-
         self.multimodal_decoder, multimodal_cfg = _build_input_dependent_text_tower(
             embed_dim, multimodal_cfg, quick_gelu, cast_dtype
         )
 
+        text = _build_input_dependent_text_tower(multimodal_cfg.width, text_cfg, quick_gelu, cast_dtype, multimodal=False)
+        self.text = text
```
iejMac, line 91: self.text = _build...

src/open_clip/coca_model.py
```diff
@@ -76,62 +77,54 @@ def __init__(
...
         self.to_logits = nn.Sequential(
-            norm_layer(embed_dim), nn.Linear(embed_dim, self.vocab_size, bias=False)
+            norm_layer(multimodal_cfg.width), nn.Linear(multimodal_cfg.width, vocab_size, bias=False)
         )
 
+        self.to_img_latent = nn.Linear(multimodal_cfg.width, multimodal_cfg.latent_dim, bias=False)
+        self.to_txt_latent = nn.Linear(multimodal_cfg.width, multimodal_cfg.latent_dim, bias=False)
         # tie embedding weights and projection
-        self.to_logits[-1].weight = self.token_embedding.weight
+        # self.to_logits[-1].weight = self.text.token_embedding.weight
```
iejMac, line 113: clean up if not needed

iejMac:
Hey, nice work! Left some very minor comments but I still need to look at the HF stuff in more detail. I'll do that later

gpucce:
> Hey, nice work! Left some very minor comments but I still need to look at the HF stuff in more detail. I'll do that later

Thanks for looking into it! For HF I didn't add a new Pooler but can do it if you think it is better

iejMac:
For HF stuff from my quick look I have 2 concerns:
1. Do all HF models have resize_token_embeddings?
2. What is the point of embed_cls? We already have a CLS pooler: https://github.com/gpucce/open_clip/blob/0e46a0199d6aaae69dc5c5089a62e8fe3b9d26ab/src/open_clip/hf_model.py#L65

gpucce:
> For HF stuff from my quick look I have 2 concerns:
> 
> 1. Do all HF models have resize_token_embeddings?
> 
> 2. What is the point of embed_cls? We already have a CLS pooler: https://github.com/gpucce/open_clip/blob/0e46a0199d6aaae69dc5c5089a62e8fe3b9d26ab/src/open_clip/hf_model.py#L65

1. Should be a method of PretrainedTransformer https://huggingface.co/docs/transformers/main/en/main_classes/model#transformers.PreTrainedModel.resize_token_embeddings

2. The reason would be to have the model CLS token and the one used for contrastive loss to be different. And adding this through the HF tokenizer seems more difficult to make stable.

iejMac:
@gpucce Re point 2: I think we actually want the CLS tokens to be the same. The idea would be to tune the CLS output embedding so that it's useful contrastively vs. just learning a completely new one. I realize this raises the issue of transformers that don't use CLS pooling but I don't see a problem with diverging from CoCa a bit here i.e. allowing for different pooling methods (like averaging) and using that as the text embedding for contrastive learning. To stay consistent with the paper we can make CLS pooling the pooling method for the default configuration.

iejMac:
Let me know what you think and also if you're busy. I can get to it myself actually in case you are

gpucce:
> Let me know what you think and also if you're busy. I can get to it myself actually in case you are

Hi, I am getting to it right now and I agree with you. Another issue might then be that what is fed to the generative head won't have the  seq <end_of_text> look

gpucce:
@iejMac would it be bad to add the `output_tokens` option to the poolers themselves?

iejMac:
Why do we want to add it to the poolers? Can't we just use the poolers to get the pooled embedding but also return the tokens on the side?

gpucce:
> Why do we want to add it to the poolers? Can't we just use the poolers to get the pooled embedding but also return the tokens on the side?

I was thinking because with CLS pooling the CLS embedding should not be in the tokens while with Mean/Max maybe yes? Can also pick from the second on for all of them

iejMac:
Ah do you mean that for CLS pooling the tokens are x[1:] whereas for mean/max it's just x? If so I still think maybe jej ust keep poolers doing only pooling and then we can add some condition where it's like
```python
if type(tokenizer) == ClsPooler:
    toks = model.hidden_state[1:]
```

or something like that, I forgot the exact syntax

src/open_clip/hf_model.py
```diff
@@ -132,12 +131,21 @@ def __init__(
...
+    def forward(self, x: TensorType, output_tokens=False) -> TensorType:
         attn_mask = (x != self.config.pad_token_id).long()
         out = self.transformer(input_ids=x, attention_mask=attn_mask)
         pooled_out = self.pooler(out, attn_mask)
+        projected = self.proj(pooled_out)
 
-        return self.proj(pooled_out)
+        if output_tokens:
+            tokens = self.proj(
+                out.last_hidden_state[:, 1:, :]
```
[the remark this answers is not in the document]
iejMac: might make code a bit ugly but at least let's save ourselves some really confusing bugs in the form of just not taking the CLS token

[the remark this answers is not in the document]
iejMac: current version doesn't fix it either. That's why I said it's going to be a little ugly because we need to do tokens = everything but the CLS token at the given position. I'm fine with only supporting the 2 most common options: first position and last position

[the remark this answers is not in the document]
iejMac: so maybe instead of slicing you can do something like tok = all_tokens and then drop cls_token_position

src/open_clip/hf_model.py
```diff
@@ -132,12 +131,21 @@ def __init__(
...
+        projected = self.proj(pooled_out)
 
-        return self.proj(pooled_out)
+        if output_tokens:
+            tokens = self.proj(
+                out.last_hidden_state[:, self.pooler.cls_token_position + 1:, :] 
+                if type(self.pooler) == ClsPooler 
+                else out.last_hidden_state
+            )
+            return projected, tokens
```
iejMac, line 146: why don't we just do an output dict if we want to add things like this

src/open_clip/hf_model.py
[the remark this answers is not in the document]
gpucce:
```
        if output_tokens:
            seq_len = out.last_hidden_state.shape[1]
            tokens = self.proj(
                out.last_hidden_state[:, torch.arange(seq_len) != self.pooler.cls_token_position, :] 
                if type(self.pooler) == ClsPooler 
                else out.last_hidden_state
            )
            return projected, tokens
```

src/open_clip/hf_model.py
gpucce: would need to add it to the non HF model too if that is fine I do it

src/open_clip/hf_model.py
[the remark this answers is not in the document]
gpucce: is this ok?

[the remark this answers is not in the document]
iejMac: I think so, if it works ofc.

src/open_clip/hf_model.py
iejMac: ah I see... hmm let me think

iejMac: let's keep it like this for now you're right

iejMac:
Hmm it would be nice to test this ClsPooler... do you know a single text encoder on huggingface that uses CLS pooling? Maybe BERT?

gpucce:
> Hmm it would be nice to test this ClsPooler... do you know a single text encoder on huggingface that uses CLS pooling? Maybe BERT?

I am trying with Roberta setting pooler_type to `cls_pooler` in the config and it seems to work , however the issue I am having now is that this way the tokens are 1 shorter than the labels and so the captioning loss errors, trying to find a simple way to adjust for it

iejMac:
yeah I guess with CLS pooling you just need to adjust labels to also be shorter

iejMac:
Ok so let's summarize this PR a bit. The tasks (from the TODO list in the coca -> main PR) that this PR resolves are:
1. Make Coca use the visual encoder and text encoder of openclip fully
2. Revert changes to Attention / CustomResidualAttentionBlock, we should avoid using them with the CoCa model
3. Move all models to output dicts instead of tuples, makes it more flexible pass output through optional loss / filters w/o brittle indexing and tuple len checks that are bound to fail with one more addition. Could make the dict output optional at construction time to prevent possible breaks for other downstream users.

Some notes:
1. Cleans up the copied inference code from the coca branch and also adds compatibility with our HFTextEncoder class. We allow making CoCa architectures that differ from the original suggested CLS pooling by just using text encoders that have different pooling methods.
2. Adds the option to use CLS pooling for the TextTransformer class

iejMac:
@gpucce any notes from you?

gpucce:
> @gpucce any notes from you?

Nothing more, writing down what I think I missed:
 - make model dual tower, I think I misunderstood the point and the idea would be to remove `self.multimodal_decoder` and move it within `self.text` (maybe)
 - log clip/caption losses separately

gpucce:
Oh and last thing this breaks `.generate` should be easy to fix but since there is also another PR working on it I would keep it as is for now.

iejMac:
https://wandb.ai/iejmac/open-clip/reports/CoCa-ViT-B-32-v2--VmlldzozMjM0Nzg1 running this for a few epochs to see if the loss roughly matches up

iejMac:
@gpucce something looks wrong, new loss is much higher
Screen Shot 2022-12-29 at 7 31 08 PM

iejMac:
@gpucce do you think we could log the losses separately in this PR? It would be much nicer for testing

gpucce:
@iejMac sure I will add it here shortly, I have an idea what the issue is too, later I will try with it a bit and late write here what I find

gpucce:
@iejMac this should log them separately

iejMac:
@gpucce we should probably also log the total loss since it's weighted so that might contain important info

gpucce:
@iejMac my idea about the issue seems like it's wrong, another thing could be that since pad tokens are now ignored in the loss it looks higher. Tomorrow I will try and check for other things, I saw your run seems to be doing ok, if you get any evaluation on a checkpoint please share, and sorry for taking a lot of time today.

rom1504:
it's quite possible the loss will go down a little later

iejMac:
@gpucce if you look at the Wandb it looks like this PR consistently achieves much higher loss and slightly lower zero shot imagenet performance than our previous version (in the coca branch). Looking over the code again, do you have any ideas if the is expected and if not, what could be wrong? It looks like the captioning loss is pretty high, too bad we can't compare to previous run.

gpucce:
@iejMac sorry I didn't have time sooner:
- higher loss is due to now ignoring pad_tokens and indeed performance is lower but close to the same
- the small decrease in performance is my mistake, I added a projection that doesn't make sense and I am confident a small commit should fix it

Before fixing this and making the model not loadable again, I added a PR to CLIP_benchmark https://github.com/LAION-AI/CLIP_benchmark/pull/63 that works with both the coca branch and this one at this stage, I tested the old checkpoint and I get scores like the following 

`Bleu_1: 0.225, Bleu_2: 0.086, Bleu_3: 0.037, Bleu_4: 0.017, METEOR: 0.064, ROUGE_L: 0.169, CIDEr: 0.089, SPICE: 0.028` 

This is not really a good validation but should be enough to check if the model generative part is training alright indeed running on an untrained model gets all zeros. If you have time could try and run this with the old checkpoint as I did and with the one from the shorter run that finished a couple days ago?

iejMac:
@gpucce no worries. I uploaded 2 comparison checkpoints to hf here - https://huggingface.co/iejMac/CoCa_tests/tree/main/checkpoints

epoch_53_new.pt is from latest run
epoch_53_old.pt is from our full run on the coca branch

gpucce:
> @gpucce no worries. I uploaded 2 comparison checkpoints to hf here - https://huggingface.co/iejMac/CoCa_tests/tree/main/checkpoints
> 
> epoch_53_new.pt is from latest run epoch_53_old.pt is from our full run on the coca branch

@iejMac here are the scores

|         |   Bleu_1 |   Bleu_2 |   Bleu_3 |   Bleu_4 |   CIDEr |   METEOR |   ROUGE_L |   SPICE |
|:--------|---------:|---------:|---------:|---------:|--------:|---------:|----------:|--------:|
| epoch_53_new |   0.1789 |   0.0363 |   0.0056 |   0      |  0.0342 |   0.0742 |    0.1605 |  0.0281 |
| epoch_53_old |   0.2004 |   0.0696 |   0.0284 |   0.0115 |  0.0704 |   0.0573 |    0.1533 |  0.0233 |

This ones are a bit worse too but probably for the same reason as the CLIP one is will push the fix in a few hours

iejMac:
@gpucce let me know when you're happy with the fixes

gpucce:
@iejMac hopefully it is fine now, if you can try running it so we see what happens it would be very nice

iejMac:
@gpucce I hope you don't mind me asking but what is this new CoCaEncoderDecoder class? I thought it was only going to be a small change. Why do you think all of this is necessary?

gpucce:
@iejMac not at all, the new class is just to have only `.text` instead of both `.text` and `.multimodal_decoder`. 

More in general the reason the change is larger than I thought is that I would have had to use even more parts of the `TextTransformer` and `VisionTransforner` from "outside" with things like `if hasattr(self.visual, "smth") ...` and it was becoming too much so I moved some of the logic as options inside those,

rom1504:
new run at https://wandb.ai/iejmac/open-clip/reports/CoCa-ViT-B-32-v2--VmlldzozMjM0Nzg1

merging here

let's continue at https://github.com/mlfoundations/open_clip/pull/308 to make sure we addressed all the points

Soonhwan-Kwon:
> new run at https://wandb.ai/iejmac/open-clip/reports/CoCa-ViT-B-32-v2--VmlldzozMjM0Nzg1
> 
> merging here
> 
> let's continue at #308 to make sure we addressed all the points

It seems that epoch_53_new comes from older version, are new run checkpoints  from latest model architecture?

PR #323, opened 2022-12-27
Add TextTextCLIP

lingjzhu:
This pull request adds TextTextCLIP (CLIP-like text-to-text contrastive retrieval model) to the main branch. 
It is still a work in progress. 

Tasks
- [X] Add a config file for TextTextCLIP
- [X] Add TextTextCLIP in `model.py`
- [X] Modify `factory.py` to load model
- [X] Modify  `data.py` to load text data
- [ ] Modify `main.py' to  train TextTextCLIP
- [X] Test loading TextTextCLIP 
- [ ] Test loading text-pair data. 
- [ ] Test dummy training

### DIFF
--- a/src/open_clip/factory.py
+++ b/src/open_clip/factory.py
@@ -228,7 +228,7 @@ def create_model_and_transforms(
         pretrained_image=pretrained_image,
         pretrained_hf=pretrained_hf,
         cache_dir=cache_dir,
-        text_to_text: Optional[bool] = False，
+        text_to_text，
     )
 
     if not text_to_text:
@@ -249,7 +249,7 @@ def create_model_and_transforms(
     else:
         preprocess_val = None
         preprocess_train = None
-        
+
     return model, preprocess_train, preprocess_val
 
 
--- a/src/open_clip/factory.py
+++ b/src/open_clip/factory.py
@@ -10,7 +10,7 @@ from typing import Optional, Tuple, Union
 import torch
 
 from .constants import OPENAI_DATASET_MEAN, OPENAI_DATASET_STD
-from .model import CLIP, CustomTextCLIP, convert_weights_to_lp, convert_to_custom_text_state_dict,\
+from .model import CLIP, CustomTextCLIP, TextTextCLIP, convert_weights_to_lp, convert_to_custom_text_state_dict,\
     resize_pos_embed, get_cast_dtype
 from .openai import load_openai_model
 from .pretrained import is_pretrained_cfg, get_pretrained_cfg, download_pretrained, list_pretrained_tags_by_model
@@ -109,6 +109,7 @@ def create_model(
         pretrained_image: bool = False,
         pretrained_hf: bool = True,
         cache_dir: Optional[str] = None,
+        text_to_text: Optional[bool] = False，
 ):
     model_name = model_name.replace('/', '-')  # for callers using old naming with / in ViT names
     if isinstance(device, str):
@@ -148,13 +149,22 @@ def create_model(
 
         cast_dtype = get_cast_dtype(precision)
         custom_text = model_cfg.pop('custom_text', False) or force_custom_text or ('hf_model_name' in model_cfg.get('text_cfg', {}))
-
-        if custom_text:
-            if 'hf_model_name' in model_cfg.get('text_cfg', {}):
-                model_cfg['text_cfg']['hf_model_pretrained'] = pretrained_hf
-            model = CustomTextCLIP(**model_cfg, cast_dtype=cast_dtype)
+        
+        # switch to TextTextCLIP
+        if text_to_text:
+            if 'hf_model_name' in model_cfg.get('doc_cfg', {}):
+                model_cfg['doc_cfg']['hf_model_pretrained'] = pretrained_hf
+            if 'hf_model_name' in model_cfg.get('query_cfg', {}):
+                model_cfg['query_cfg']['hf_model_pretrained'] = pretrained_hf
+                
+                model = TextTextCLIP(**model_cfg, cast_dtype=cast_dtype)
         else:
-            model = CLIP(**model_cfg, cast_dtype=cast_dtype)
+            if custom_text:
+                if 'hf_model_name' in model_cfg.get('text_cfg', {}):
+                    model_cfg['text_cfg']['hf_model_pretrained'] = pretrained_hf
+                model = CustomTextCLIP(**model_cfg, cast_dtype=cast_dtype)
+            else:
+                model = CLIP(**model_cfg, cast_dtype=cast_dtype)
 
         pretrained_cfg = {}
         if pretrained:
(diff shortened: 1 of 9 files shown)


src/training/params.py
```diff
@@ -326,13 +326,54 @@ def parse_args(args):
         action='store_true',
         help="Freeze BatchNorm running stats in image tower for any locked layers.",
     )
+    parser.add_argument(
+        "--lock-doc",
```
rom1504, line 330: what about a "lock-tower-1" and "lock-tower-2" param instead ?
query/doc is not the only thing

lingjzhu: Sure. I will rename these variables immediately. Do you think we should also rename variables like `image_features` in `train.py` and `main.py'? Will renaming variables cause confusion for the original text-image model? Currently `train.py` is fully compatible with TextTextCLIP. No modification other than renaming variables is needed.

lingjzhu: I have tested TextTextCLIP locally and it seems to work smoothly. But the script fails some tests here. I will make adjustments.

src/training/params.py
```diff
@@ -326,13 +338,54 @@ def parse_args(args):
         action='store_true',
         help="Freeze BatchNorm running stats in image tower for any locked layers.",
     )
+    parser.add_argument(
```
rom1504, line 341: wonder if this can be reconciled with the normal locking params

src/training/data.py
```diff
@@ -27,6 +27,54 @@
     hvd = None
 
 
+class TextPairDataset(Dataset):
+    def __init__(self, input_filename, text_a_key, text_b_key, tokenizer=None):
+        logging.debug(f'Loading parquet data from {input_filename}.')
+        df = pd.read_parquet(input_filename)
```
rom1504, line 33: this is unlikely to scale

src/open_clip/model.py
```diff
@@ -248,6 +248,47 @@ def forward(self, image, text):
         return image_features, text_features, self.logit_scale.exp()
 
 
+
+class TextTextCLIP(nn.Module):
```
rom1504, line 252: this is completely duplicated from above class, I wonder if we could reconcile it

src/open_clip/factory.py
```diff
@@ -263,18 +285,20 @@ def create_model_from_pretrained(
...
 
-    image_mean = image_mean or getattr(model.visual, 'image_mean', None)
-    image_std = image_std or getattr(model.visual, 'image_std', None)
-    preprocess = image_transform(
-        model.visual.image_size,
-        is_train=False,
-        mean=image_mean,
-        std=image_std
-    )
+    if not text_to_text:
```
rom1504, line 294: what about checking if model.visual exists instead ?

src/open_clip/factory.py
```diff
@@ -179,9 +193,10 @@ def create_model(
         if precision in ("fp16", "bf16"):
             convert_weights_to_lp(model, dtype=torch.bfloat16 if precision == 'bf16' else torch.float16)
 
+        if not text_to_text:
```
rom1504, line 196: what about checking if model.visual exists instead ?

src/open_clip/model.py
lingjzhu: Maybe we could use this one as the general model? It's not specific about modality. We can refer to `image_features` as `features_a` and `text_features` as `features_b`.

src/open_clip/factory.py
lingjzhu: If we merge `TextTextCLIP` to `CustomCLIP` to get a more general model, then `model.visual` might not exist at all?

src/training/data.py
lingjzhu: Right. I will update this. Do you have any suggestions?

rom1504:
Can you rebase on master ?

lingjzhu:
> Can you rebase on master ?

Done! I've been setting up training in the computing cluster and fixing typos here and there. The code is mostly OK now. It should be ready once training has been successfully completed. I will post `wandb` logs for model training soon.

lingjzhu:
@rom1504 Here is the latest code for TextTextCLIP. It is tested on the stability cluster and the evaluation code is also included. I have added an example script at `docs/script_examples/text_example.sh`. Let me know if it is OK to merge or more revisions are needed. Thank you!

rom1504:
This would really benefits from reviews @rwightman @mitchellnw @iejMac 
It was trained quite decently
Getting it in a good enough state and merged would go a long way to expend the applicability space of openclip

rwightman:
@rom1504 k, will try and look at it soon

src/open_clip/loss.py
```diff
@@ -99,33 +99,44 @@ def get_ground_truth(self, device, num_logits) -> torch.Tensor:
...
+            logits_per_feature_b = logit_scale * features_b @ features_a.T
+
+       
+        return logits_per_feature_a, logits_per_feature_b
+
+
+    def forward(self, image_features=None, text_features=None, logit_scale=None, text_a_features=None, text_b_features=None, output_dict=False):
         
-        return logits_per_image, logits_per_text
+        if image_features is not None and text_features is not None:
```
daniel-z-kaplan, line 125: This section needs a check for the case of both image/text are none, since that's something new that can occur now I think?

src/open_clip/loss.py
```diff
@@ -99,33 +99,44 @@ def get_ground_truth(self, device, num_logits) -> torch.Tensor:
...
-            logits_per_image = logit_scale * image_features @ text_features.T
-            logits_per_text = logit_scale * text_features @ image_features.T
+            logits_per_feature_a = logit_scale * features_a @ features_b.T
+            logits_per_feature_b = logit_scale * features_b @ features_a.T
+
+       
+        return logits_per_feature_a, logits_per_feature_b
+
+
+    def forward(self, image_features=None, text_features=None, logit_scale=None, text_a_features=None, text_b_features=None, output_dict=False):
```
daniel-z-kaplan, line 124: This section feels a little awkward, maybe it should just always take features_a and features_b, not optional, like the prior code?
Same with logit_scale, why is that optional now?

rwightman:
overall things look pretty good, I'm trying to get over a mental block re the loss naming, I reliaze why the feature_a/b changes were made to the loss but I feel it harms ease of understanding re the  most common use case, especially for newcomers to the core loss fn... hmm hnm

src/open_clip/loss.py
lingjzhu: Thank you for your comments! This is an artifact of trying to reconcile the original CLIP and the text model. But I agree with you that this is a bit awkward. I could add a loss wrapper for the text model. What do you think?

By the way, do you have any ideas as to the naming of the text model? TextTextCLIP sounds a bit redundant to me. How about CTTP? CLLP? LATTE (contrastive LAnguage-To-TExt pretraining)? CTP (contrastive text pretraining)? Do you have a better suggestions?

Issue #278, opened 2022-12-07
braceexpand complex urls leads to memory explosion


zhouyizhuang-megvii:
Hi,

I encounter a corner case that may lead to memory explosion when pass a complex webdataset url pattern to `args.train_data` (i.e. `input_shards`) .

In my case, my `args.train_data` is `s3://collect-22040715001-data/laion400m_part0/{00000..09999}.tar::s3://collect-22040715001-data/laion400m_part1/{10000..19999}.tar::s3://collect-22040715001-data/laion400m_part2/{20000..29999}.tar::s3://collect-22040715001-data/laion400m_part3/{30000..41455}.tar` since my laion400m is chunked into 4 directories. I use `::` to concat them, which can be correctly resolved in webdataset (i.e. `urls = wds.shardlists.expand_urls(urls)`)

However, when calulating the chunk size when preparing training webdataset, this url pattern is naively expanded, causing approximately 10000^4 = 10^16 possible urls, the program thus stucking and memory blowing up. 

https://github.com/mlfoundations/open_clip/blob/45278d27fd6da041b184dea84476f4924947ffad/src/training/data.py#L78-L79

I think a quick solution is to expand any complex urls with the simple `wds.shardlists.expand_urls` tool webdataset provided, in case any unlucky user like me face the same issue in the future.

Thank you!

gabrielilharco:
There's a quick fix at https://github.com/mlfoundations/open_clip/pull/336

PR #336, opened 2023-01-04
Fix braceexpand memory explosion for complex urls

gabrielilharco:
Currently handling complex webdataset url patterns in `args.train_data` can lead to a unnecessary memory explosion, when using `::` to concatenate multiple data sources. This can be correctly parsed by webdataset, using `wds.shardlists.expand_urls(urls))`.

See Issue https://github.com/mlfoundations/open_clip/issues/278.

### DIFF
--- a/src/training/data.py
+++ b/src/training/data.py
@@ -9,7 +9,6 @@ import time
 from dataclasses import dataclass
 from multiprocessing import Value
 
-import braceexpand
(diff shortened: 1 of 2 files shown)


rom1504:
Could you add an unit test to make sure it works ?

rom1504:
Lgtm

PR #341, opened 2023-01-06
add beam search,fix image_latent,image_emb calculated once

Soonhwan-Kwon:
Proposal to add beam search for coca.
fix image_latents and image_tokens are calculated once instead of every seq_len.
generate time reduced, 2.5967->1.8194, generate_beamsearch reduced(5 images) 53.1091->30.3831

- [x] beam search w/o past key values
- [ ] beam search w/ past key values

rebased version of PR: https://github.com/mlfoundations/open_clip/pull/320


Soonhwan-Kwon:
@rom1504 fixed test errors

gpucce:
@Soonhwan-Kwon I have a question as I am not sure, having the `past_key_values` in the loss, is useful in general or is it an option if one wants to compute the loss with respect to the previously generated tokens instead of the ground truth?

Soonhwan-Kwon:
> @Soonhwan-Kwon I have a question as I am not sure, having the `past_key_values` in the loss, is useful in general or is it an option if one wants to compute the loss with respect to the previously generated tokens instead of the ground truth?

It is natural to have the question about it, and it won't do any role to calculate loss. It is mere workaround not to make changes for forward method of model.py. And I'll remove changes related to past_key_values as you mentioned.

rom1504:
@Soonhwan-Kwon if you want to finish this, it'd be appreciated

Soonhwan-Kwon:
> 



> @Soonhwan-Kwon if you want to finish this, it'd be appreciated

If you mean finish this PR for merge, I think beam search w/ past key values can be done in later, because it can affect forward of CoCa. Benchmark for MS COCO captions can be done in several minutes or less w/o past key values for now.

rom1504:
ok let's focus on fixing things so the CI passes for now

Soonhwan-Kwon:
@rom1504 I fixed to pass the CI and wrote comments in coca PR  also.

Issue #340, opened 2023-01-05
Wandb runs combined when using resume


afang-story:
When making multiple runs with resume as latest, they get combined into the same run on wandb and the name gets changed to the latest one, even though the names are unique.
According to https://docs.wandb.ai/ref/python/init this is because resuming needs a unique id or else it will overwrite the old run.
The solution to this problem would be to add an id to the wandb init call main.py.

rom1504:
The intended behavior is to have a single run.
Would you like an option to have multiple runs ?

gabrielilharco:
It's nice to have support for running multiple runs in parallel imo. Would adding `id=args.name` to https://github.com/mlfoundations/open_clip/blob/main/src/training/main.py#L292 fix this?

mitchellnw:
that would probably do it @gabrielilharco

gabrielilharco:
I'll test this and open a PR if it works

gabrielilharco:
Fixed in https://github.com/mlfoundations/open_clip/pull/344

PR #344, opened 2023-01-07
set wandb id to args.name

gabrielilharco:
Set `id` to `args.name` when initializing wandb so multiple parallel runs are not collapsed into a single one. See https://github.com/mlfoundations/open_clip/issues/340.


mitchellnw:
LGTM - @gabrielilharco is this ready to merge?

gabrielilharco:
yes, tested and working

PR #345, opened 2023-01-08
Add backward compatibility tests

gpucce:
Adding tests to check that new output logic doesn't break past output logic


gpucce:
@rom1504 @rwightman @iejMac 

Making the models jit compilable took more changes than I thought and I don't know much about torchscript in general, this should add a test that verifies it and I think coca currently can't be jit compiled due to a bug, I opened an issue to understand if I am missing smth https://github.com/pytorch/pytorch/issues/92073

Please, If you have time, have a look at how I did it because I could be smarter solutions that I am not aware of.

I also merged the branch adding generate to see if it was working with that too.

rwightman:
@gpucce @rom1504 k, more or less confirms what I was expecting, it's a mess and changes broke torchscript. not quite sure what the best approach is, see some things that could be improved but overall the goal of maintining a decent amount of modelling bwd compat, plus supporting varying output types (dict vs tensors vs tuple) is going to make torchscript a headache. Hmm

rom1504:
What I see here
* Beam generation code is huge and can most likely be improved 
* test is not done right, add cases to the pytest parametrize, no need to duplicate it
* You need some options to make jit fast (see discussion in other PR)
* in current state i don't see a lot of changes to keep jit compatibility, is that expected?

gpucce:
This continues in https://github.com/mlfoundations/open_clip/pull/355

PR #355, opened 2023-01-13
Add tests bis

gpucce:
This should make clip and coca compatible with jit although for coca this needs to be fixed first https://github.com/pytorch/pytorch/issues/92073

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,7 @@ jobs:
           --group ${{ matrix.job }} \
           -m regression_test \
           tests \
-          | head -n -2 | grep -Po 'test_inference_with_data\[\K[^]]*(?=])' \
+          | head -n -2 | grep -Po 'test_inference_with_data\[\K[^]]*(?=-False]|-True])' \
           > models_gh_runner.txt
         if [ -n "${{ inputs.manual_revision_reference }}" ]; then
           REVISION_REFERENCE=${{ inputs.manual_revision_reference }}
--- a/src/open_clip/coca_model.py
+++ b/src/open_clip/coca_model.py
@@ -126,13 +126,13 @@ class CoCa(nn.Module):
         self.text.set_grad_checkpointing(enable)
 
     def _encode_image(self, images, normalize=True):
-        image_latent, tokens_embs = self.visual.encode_with_tokens(images)
+        image_latent, tokens_embs = self.visual(images)
         image_latent = F.normalize(image_latent, dim=-1) if normalize else image_latent
         return image_latent, tokens_embs
 
     def _encode_text(self, text, normalize=True):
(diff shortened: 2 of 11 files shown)


gpucce:
@rom1504 this is the cleaner version of making everything jit compilable, if you can have a look at how it is done now I would appreciate it, this should make both everything backward compatible and add the possibility to output dicts, for coca itself it can't be tested yet with jit until https://github.com/pytorch/pytorch/issues/92073 is fixed.

Also for now I only test ViT-B-32 because I don't know which model you want to test with jit (HF seems to not support it independently of open_clip)

gpucce:
@rom1504 this now works but probably breaks someone else's code if they where using CLIP as part of a larger model, however I don't know how to get around that while adding the `output_dict` option

rom1504:
encode_with_tokens is the main change here right?

Basically you're trying to keep exactly the same signature as before, no new optional param ?

It's not too bad but also it makes the code a little uglier

rom1504:
> breaks someone else's code if they where using CLIP

If that's true, then what's the gain of this PR changes ?

gpucce:
@rom1504 after your comment (thanks again) this is how it should work now:
- all models except for clip must output dict and this is tested in `test_inference`;
- clip is allowed to output a tuple as before and it is converted to a dict in `train.py`, this is done only if the model is `CLIP` or `CustomTextCLIP` and still in `test_inference` it is checked that these two models output the tuple as it used to be and coherently with how they output dicts;
- whether CLIP outputs dict or tuple is decided by an `output_dict` param in the config, which defaults to `False` so everything stays as it was for existing things;
- everything is jit compilable and we test with a wrapper for clip and check that it is itself compilable in jit both if clip outputs dicts or tuples;
- all the tests are in `test_inference` and for now I only jit test one model `ViT-B-32` because I don't know which one is best to check;
- for `output_with_tokens` the idea is indeed to have no changes at all in how text/vision-transformer works, because otherwise the number of params to specify becomes very larger and a bit confusing, I can change it if it is better otherwise;

gpucce:
@rom1504 when you have time for a review this should be ready, if it is still missing smth I will add it or fix it

rom1504:
thanks for the work and sorry for delay

I'm wondering if we really need to duplicate forward methods

is it only for jit purposes ?

gpucce:
@rom1504 yes they are only for jit, now I removed them to use parameters from the configs, I don't know what the best approach is.

Now the forwards should be unchanged except for the option to `output_tokens` which taken from an attribute of the model and all seems to be ok with jit and tested to be ok with it.

Inside coca model I kept the `_encode_image` and `_encode_text` methods next to `encode_image` and `encode_text` as they make it easier to do things coherently with clip with less changes

rom1504:
Thanks

Let's continue at main branch

We're trying to merge soon

PR #360, opened 2023-01-21
Example slurm script

mitchellnw:
Using remote sync and resume latest.

### DIFF
--- a/docs/script_examples/stability_example.sh
+++ b/docs/script_examples/stability_example.sh
@@ -27,7 +27,7 @@ export PYTHONPATH="$PYTHONPATH:/admin/home-mitchellw/open_clip/src"
 
 EXP_NAME="test-B-32-laion5b-lr1e-3-bs90k"
 
-/opt/slurm/sbin/srun --comment laion --cpu_bind=v --accel-bind=gn python -m training.main \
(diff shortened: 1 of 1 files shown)


mitchellnw:
Note this actually won't work currently on stability due to the `srun` issue. The workaround I used is to add these lines of code after `args` are obtained in the `main` function:

```
    if args.train_data is not None and args.train_data.startswith('s3'):
        args.train_data = f"pipe:aws s3 cp {args.train_data} -"
```

docs/script_examples/stability_example.sh
```diff
@@ -0,0 +1,53 @@
...
+#SBATCH --open-mode=append
+#SBATCH --exclusive
+
+module load openmpi
+module load cuda/11.7
+
+export PYTHONFAULTHANDLER=1
+export CUDA_LAUNCH_BLOCKING=0
+export HOSTNAMES=`scontrol show hostnames "$SLURM_JOB_NODELIST"`
+export MASTER_ADDR=$(scontrol show hostnames "$SLURM_JOB_NODELIST" | head -n 1)
```
rom1504, line 18: this should be replaced by slurm env var

rom1504:
good enough

PR #366, opened 2023-01-22
do same thing for _encode_image

iejMac:


rom1504:
FAILED tests/test_training_simple.py::test_training_coca - AttributeError: 'CoCaEncoderDecoder' object has no attribute 'output_tokens'

src/open_clip/coca_model.py
```diff
@@ -121,30 +121,25 @@ def set_grad_checkpointing(self, enable=True):
         self.visual.set_grad_checkpointing(enable)
         self.text.set_grad_checkpointing(enable)
 
-    def _encode_image(self, images, normalize=True):
+    def encode_image(self, images, normalize=True):
         image_latent, tokens_embs = self.visual(images)
         image_latent = F.normalize(image_latent, dim=-1) if normalize else image_latent
-        return image_latent, tokens_embs
+        if self.visual.output_tokens:
```
rom1504, line 127: what is this trying to do ?

why don't we have a output_tokens option in encode_image ?

iejMac: it's trying to see if the visual encoder is configured to output tokens, if so, output them, otherwise don't

PR #368, opened 2023-01-22
remove output_dict argument

gpucce:
@rom1504 I left an unused argument


rom1504:
This argument is used

gpucce:
> This argument is used

Sorry, to make it work with torchscript I had to use a model attribute instead of the function argument but along the changes I had not removed it from the method, although I was overriding it in the body

PR #373, opened 2023-01-23
better torchscript and small test refactor

gpucce:
@rwightman @iejMac @rom1504 

@rwightman was saying that booleans should be ok for torchscript, if I understand they are but to avoid checking all branches output the same thing one needs to make them constants https://pytorch.org/docs/stable/jit_language_reference.html#python-defined-constants 

In this PR I annotate them as constants and it seems to work. Let me know if this is ok and thank you for all the reviews

### DIFF
--- a/src/open_clip/hf_model.py
+++ b/src/open_clip/hf_model.py
@@ -81,7 +81,7 @@ class ClsPooler(nn.Module):
 
 class HFTextEncoder(nn.Module):
     """HuggingFace model adapter"""
-
+    output_tokens: torch.jit.Final[bool]
     def __init__(
             self,
             model_name_or_path: str,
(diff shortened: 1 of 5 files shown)


rwightman:
@gpucce thanks for looking at this, constant works too I guess, if it's annotated constant, when it compiles it will only cover one of the paths, but that means a model exported via torchscript would be locked on that one path, probably fine here... but for the base model I checked this and it works (with a non const bool), so that compiles the conditional and leave either path as possible at runtime

```
    def forward(self, image, text) -> Union[Dict[str, torch.Tensor], Tuple[torch.Tensor, torch.Tensor, torch.Tensor]]:
        if self.output_dict:
            return {
                "image_features":image_features, 
                "text_features":text_features,
                "logit_scale":self.logit_scale.exp()
            }
        return image_features, text_features, self.logit_scale.exp()
```

gpucce:
I was using this but then if I understand it would be very hard to compile a model with clip ad a component, that's why I didn't use it in the end. I will change it to this!

rwightman:
@gpucce let's get opinion from others, the const approach you've done here could be better for integration as you point out, don't see a case where someone would want to change it after scripting for a given use case?

PR #374, opened 2023-01-23
rename timm models in tests

gpucce:
@lopho @rwightman @rom1504 

This models with names starting with "timm" don't exist anymore but the equivalent ones appear to still be unavailable in timm so I rename them in the tests too

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -13,10 +13,10 @@ models_to_test = set(open_clip.list_models())
 models_to_test = models_to_test.difference({
         # not available with timm yet
         # see https://github.com/mlfoundations/open_clip/issues/219
-        'timm-convnext_xlarge',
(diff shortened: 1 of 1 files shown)


rwightman:
@gpucce thanks! .. I guess a question there is, why didn't that make the tests fail?

gpucce:
I think the data for those models is not generated so they are skipped, but I am not totally sure

lopho:
Regression tests on the CI generate test data from previous commits (v2.8.0 atm).
Since its still generating that data using the old model names as present at that time, the testing data name will be different.
Testing is done at HEAD (against data generated at aforementioned tag).
As such the test is skipped because it will not find data according to the config name used to generate the data.
I've done it this way to ensure only models that are present at both points in the history are considered for regression testing.

Issue #370, opened 2023-01-22
support for distillation


mitchellnw:
consider supporting distilling existing large clips into smaller ones

rom1504:
https://github.com/iejMac/encoder-distill by @iejMac has some code for this

mitchellnw:
ready to merge i think https://github.com/mlfoundations/open_clip/pull/399

PR #386, opened 2023-01-28
[WIP] Testing out supporting distillation in open_clip

mitchellnw:
Can't support every feature so we may not want this, but just seeing how an initial experiment goes.

As per https://github.com/mlfoundations/open_clip/issues/370

This PR tries an alternative approach to @iejMac in https://github.com/iejMac/encoder-distill because it does not use MSE and instead just computes logits with both the student and teacher and does standard distillation loss


mitchellnw:
All goes well distill OAI L/14 into B/32 will take 2.8 days https://wandb.ai/dogml/open_clip6/runs/distill_l14_32b_laion2b_v2

mitchellnw:
final distillation number is 67.47% for B/32 (when distilling from L/14), so +4pp

mitchellnw:
This got too confusing and I'm going to start again from scratch.

Issue #390, opened 2023-01-29
coca follow ups


rom1504:
* [x] make B/32 checkpoint compatible with current code
* [x] release L/14 checkpoint
* [x] remove support for non dict mode in train code https://github.com/mlfoundations/open_clip/pull/308#discussion_r1089834375
* [x] add beam search generate https://github.com/mlfoundations/open_clip/pull/341
* [x] add in readme how to generate with a pretrained model
* [ ] Add instructions for captioning eval in readme  https://github.com/LAION-AI/CLIP_benchmark/pull/63
* [x] (Optional) provide a coca fine tuned for captioning on coco
* [x] Release post mentioning training runs, how to train, how to evaluate, results, how to generate 
* [ ] (Optional) past key for speed up

rwightman:
@rom1504 that's not 'remove an option to not output a dict', that was the whole point on the model side so other users of the model don't break, it's 'remove support for non-dict mode in train code'

rom1504:
Alright

vturrisi:
Hey! Thanks for the great effort in making CoCa available.

I was generating captions with CoCa and they work fine with the new beam search implementation. However, I wanted to do top-p/nucleus sampling generation as in BLIP and for that there are a couple of issues with the current implementation. There's no way to enforce minimal number of tokens and the end of sequence token is ignored, so the model will keep generating until the sequence length. I did some quick fixes on my side but since I'm not super familiar with huggingface/openclip that is for sure suboptimal.

I can share more info if needed :)

gpucce:
hi @vturrisi I opened a PR here https://github.com/mlfoundations/open_clip/pull/402 can you try it and see if this is what you expected?

rom1504:
Done. Last point tracked there https://github.com/mlfoundations/open_clip/issues/409

lucidrains:
@gpucce thank you for your amazing work Giovanni!

ludwigschmidt:
Indeed, awesome work here!

usuyama:
Awesome work!

alexcode4u:
@rom1504 Is there way to generate several outputs using top_k/top_p sampling for CoCa? May be there is some parameter like "num_return_sequences" in hugging face? I have not found anything like that in generate function. https://github.com/mlfoundations/open_clip/blob/main/src/open_clip/coca_model.py#L167

PR #393, opened 2023-01-29
CoCa: generate and beam search

gpucce:
This should add generate and beam_search correctly and address part of #390


rom1504:
Can you add instructions of use in readme?

README.md
```diff
@@ -285,6 +285,32 @@ Training [CoCa](https://arxiv.org/abs/2205.01917) models is enabled through spec
...
+
+To generate text with coca this should work
+
+```python
+import open_clip
+from PIL import Image
+
+model, _, transform = open_clip.create_model_and_transform(
+  model_name="coca_ViT-B-32",
+  pretrained="path/to/pretrained"
```
rom1504, line 298: Point to an existing model instead (see readme, the key is specified at the end)

gpucce:
@rom1504 this should be ok now

Issue #395, opened 2023-01-30
`is_clip` will fail if enable `torchscript`


wjfwzzc:
https://github.com/mlfoundations/open_clip/blob/191d138296d6749396b2b15e5c3c9459b05e65b3/src/training/train.py#L41-L42

`type(model)` will be `torch.jit._script.RecursiveScriptModule`, `isinstance` will also fail. A compromise solution is to check `model.original_name` when `model` is a `ScriptModule`.

rwightman:
@wjfwzzc should be a non-issue now, tx

PR #396, opened 2023-01-30
Make clip always output dict in open_clip traininig

gpucce:
This should make clip always use the output dict option when the training is done inside open_clip. 

Partially addressing #390 and fixing #395 

@rwightman


rwightman:
@gpucce thanks, rather than is running in openclip,  I think it should be `output_dict: Optional[bool] = None` ... if it's default of None, no change is made to the model default, otherwise, it will check attribute and pass through

open_clip will set to True

rohan1561:
Hey, should the forward in the line 263 of training/train.py contain the output_dict argument? It throws an error for now..

rwightman:
@rohan1561 there should be no output dict args in model fwd, don't know why that's still there, guess we missed one @gpucce

gpucce:
Hi, indeed there should not be, I will make a PR with also a test

rwightman:
@gpucce I already pushed  a quick fix, but PR for a test to cover this case would be good

PR #398, opened 2023-01-31
[WIP] Add support for dataset mixtures with different sampling weights

gabrielilharco:
When using multiple data sources (e.g. `--train-data=path/to/dir/1/{000..999}.tar::path/to/dir/2/{000..999}.tar`), allow the different sources to be sampled with different weights (e.g. by using `train-data-weights=100::1`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,18 @@ You can set this on your visual transformer config with the key `patch_dropout`.
 
 In the paper, they also finetuned without the patch dropout at the end. You can do this with the command-line argument `--force-patch-dropout 0.`
 
+#### Multiple data sources
+
+OpenCLIP supports using multiple data sources, by separating different data paths with `::`.
+For instance, to train on CC12M and on LAION, one might use `--train-data '/data/cc12m/cc12m-train-{0000..2175}.tar'::/data/LAION-400M/{00000..41455}.tar"`.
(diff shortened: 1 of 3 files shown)


gabrielilharco:
I tested these changes on a ViT-B/32 run on CC12M with BS 8k and lr 5e-4, and did not observe any impact on the standard workflow (25.04% vs 24.95% zero-shot ImageNet accuracy)

usuyama:
This looks awesome! 

With this feature, how many samples do we see in one "epoch"?
I think currently we have two ways to specify the number of training samples per epoch (1) json file for wds (2) train num samples argument.

gabrielilharco:
@usuyama, thanks! Right now this is designed to work with resampling (`--dataset-resampled`) and specifying `--train-num-samples`

rom1504:
Could you check that this change is not affecting the normal use of openclip eg training on one dataset?

For example by checking what gets out of the data loader before/after your change ?

gabrielilharco:
I checked that with two ViT-B/32 runs on CC12M with BS 8k and lr 5e-4 (one before and one after the change). The curves looked almost identical, and I did not observe any impact (25.04% vs 24.95% zero-shot ImageNet accuracy).

PR #399, opened 2023-01-31
Distillation

mitchellnw:
Still need to test, made changes again but after CoCa merge

### DIFF
--- a/README.md
+++ b/README.md
@@ -517,6 +517,12 @@ Future trained models will use nn.GELU.
 
 >>> model, train_transform, eval_transform = open_clip.create_model_and_transforms('ViT-B-32', pretrained='laion2b_s34b_b79k')
 ```
+### Model distillation
+
+You can distill from a pre-trained by using `--distill-model` and `--distill-pretrained` to specify the model you'd like to distill from.
+For instance, to distill from OpenAI ViT-L/14 use `--distill-model ViT-L-14 --distill-pretrained openai`.
+
+
 ### Gradient accumulation
 
 To simulate larger batches use `--accum-freq k`. If per gpu batch size, `--batch-size`, is `m`, then the effective batch size will be `k * m * num_gpus`.
--- a/src/open_clip/__init__.py
+++ b/src/open_clip/__init__.py
@@ -1,7 +1,7 @@
 from .constants import OPENAI_DATASET_MEAN, OPENAI_DATASET_STD
 from .factory import create_model, create_model_and_transforms, create_model_from_pretrained, get_tokenizer, create_loss
 from .factory import list_models, add_model_config, get_model_config, load_checkpoint
-from .loss import ClipLoss, CoCaLoss
+from .loss import ClipLoss, DistillClipLoss, CoCaLoss
 from .model import CLIP, CustomTextCLIP, CLIPTextCfg, CLIPVisionCfg,\
     convert_weights_to_lp, convert_weights_to_fp16, trace_model, get_cast_dtype
 from .coca_model import CoCa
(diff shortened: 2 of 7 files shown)


mitchellnw:
this is with equal weight on both losses which is also what coca does so would be good if this gives high accuracy and we can stick with it

https://wandb.ai/dogml/open_clip6/runs/distill_l14_32b_laion2b_v4

mitchellnw:
got 67.64% when distilling L/14 into B/32 with 13b samples seen so seems good to go (+ > 4%). a few things are not supported and have fixmes such as
- distillation + grad accum
- distillation + coca
but these should probably be follow up PRs.

mitchellnw:
anything required before merging? understand if no plans to merge due to too many features

rsomani95:
@mitchellnw the ImageNet top-1 reported was at the end of the full training run. I'm curious if you measured that every few epochs as well? I'm wondering what's the accuracy growth curve looks like - trying to build intuition for how many image samples one would need to be able to distill. 

On a related note, have you done any experiments distilling either of these scenarios:
1. A `timm` image encoder model that's been pretrained on ImageNet
2. (1) initialised with the _same_ text encoder as the teacher model

My intuition is that both of these scenarios would lead to faster convergence. In my use case, we need to run the image encoder much more often than the text encoder, so it's okay to have a heavier text model as described in (2)

mitchellnw:
Sorry we never ran more experiments on this, we only have the final number as reported! I would be very curious to see what happens.

rsomani95:
@mitchellnw got it, thanks for the response. Do you think those experiments make sense? I'd love your perspective on it and if you have any suggestions / gotchas to look out for. I will be running some form of these experiments, but have limited compute (3x 3090s), so any time saved with better assumptions would be very welcome :)

Specifically, I'll be training an ImageNet pretrained `ghostnet_100` variant with a pretrained `convnext_base_w` text encoder, and distilling from `convnext_base_w` for starters.

mitchellnw:
The distillation code we have right now distills in the contrastive image-text setting so I don't believe an imagenet pre-trained model will work

mitchellnw:
Unless you mean to go the other way?

rsomani95:
> The distillation code we have right now distills in the contrastive image-text setting so I don't believe an imagenet pre-trained model will work

Interesting. Do you think a randomly initialised img encoder is better to start with than one that's been pre-trained on ImageNet? Or maybe an img encoder that's been fine-tuned to be a bit more aligned with a CLIP image encoder?

More formally, I'm trying to evaluate these three strategies for the best initialisation step for the image encoder:
1. Random initialisation
2. ImageNet pretraining
3. An ImageNet model that's been distilled from the CLIP teacher model's image encoder _only_
    - Here, the idea is to take a few million (billion?) images, compute embeddings for the teacher and student image encoder, and do a simple mean squared error loss
    - If this helps, it can be a lot cheaper and faster to do as a step before doing the contrastive image-text distillation

mitchellnw:
If you're distilling into an encoder than I think starting that encoder with imagenet pre-trained will be better. But I also think that's not specific to distillation -- if you start CLIP training from a pre-trained encoder I also think that will be better than from scratch. I really don't know, though.

rsomani95:
Got it. Thanks, that's helpful. Will post any notable findings here as I go along.

abhishek0093:
Hi @mitchellnw , 
Thanks for this great addition to the repo. I was just wondering if these distilled models which you trained above are available somewhere? 
And do you think it would work if I finetune a B/16 and simulatenously distill it from L/14 on my custom dataset of ~2.5M datapoints.

PR #400, opened 2023-02-01
Add CoCa L/14 pretrained

iejMac:


rwightman:
@iejMac looks like the checkpoint in hf hub is train, it's 7+ GB... we should strip the train state and upload just a checkpoint that's the state_dict for the main model

iejMac:
@rwightman will do. Just wanted to get this up ASAP. I'll probably get to that later today

iejMac:
2.55GB now

PR #402, opened 2023-02-02
Add min_seq_len and support for eos_token_id in  `generate`

gpucce:
This should add support in generate for min_seq_len and stop ignoring eos_token_id in `generate`

### DIFF
--- a/src/open_clip/coca_model.py
+++ b/src/open_clip/coca_model.py
@@ -165,7 +165,7 @@ class CoCa(nn.Module):
             num_beam_groups=3,
             min_seq_len=5,
             stopping_criteria=None,
-            fixex_output_length=False # the output will have shape min(seq_len, max_output_len)
+            fixed_output_length=False # the output will have shape min(seq_len, max_output_len)
     ):
 
         sot_token_id = 49406 if sot_token_id is None else sot_token_id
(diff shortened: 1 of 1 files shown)


vturrisi:
@gpucce thanks the fast implementation. I think that there should also be a break on the mask, so that we stop the generation on that as well. Something like this?
```
if (cur_len >= min_seq_len and cur_len >= seq_len) or mask.all():
    break
```

vturrisi:
Also, when the number of tokens generated is the same as the max_seq_len or seq_len (not sure), I get this:
```
    sample[~mask, :] = (
RuntimeError: shape mismatch: value tensor of shape [0, 1] cannot be broadcast to indexing result of shape [2, 1]
```
You can force this error by setting min_seq_len=76, seq_len=77, max_seq_len=77.

vturrisi:
> > @gpucce thanks the fast implementation. I think that there should also be a break on the mask, so that we stop the generation on that as well. Something like this?
> > ```
> > if (cur_len >= min_seq_len and cur_len >= seq_len) or mask.all():
> >     break
> > ```
> 
> That would mean that the output tensor has a different shape than expected but can add for sure, anything else not working as you expected?

Yes, indeed the output has variable length. Maybe setting it as a flag?

src/open_clip/coca_model.py
```diff
@@ -165,6 +165,7 @@ def generate(
             num_beam_groups=3,
             min_seq_len=5,
             stopping_criteria=None,
+            fixex_output_length=False # the output will have shape min(seq_len, max_output_len)
```
vturrisi, line 168: typo on the name?

vturrisi:
Tested the latest commit. Couldn't find any issues. Thanks @gpucce :)

gpucce:
> Tested the latest commit. Couldn't find any issues. Thanks @gpucce :)

Thanks for the feedback and if there are some other things you would like to change/tweak please let me know!

vturrisi:
@gpucce was checking BLIP's nucleus sampling strategy and they also have a repetition penalty. I feel like we could use [RepetitionPenaltyLogitsProcessor](https://huggingface.co/docs/transformers/internal/generation_utils#transformers.RepetitionPenaltyLogitsProcessor) for that. Also, it would be good to unify and use Huggingface utility classes, like [TopPLogitsWarper](https://huggingface.co/docs/transformers/internal/generation_utils#transformers.TopPLogitsWarper), similarly to what was done for beam search.

vturrisi:
I think we only need to add:
```
from transformers import RepetitionPenaltyLogitsProcessor

repetition_penalty_processor = RepetitionPenaltyLogitsProcessor(
        penalty=repetition_penalty
    )

while True:
    ....
    logits = self(image, x, embed_cls=False)["logits"][:, -1]
    logits = repetition_penalty_processor(out, logits)
```

gpucce:
> I think we only need to add (for the repetition penalty):
> 
> ```
> from transformers import RepetitionPenaltyLogitsProcessor
> 
> repetition_penalty_processor = RepetitionPenaltyLogitsProcessor(
>         penalty=repetition_penalty
>     )
> 
> while True:
>     ....
>     logits = self(image, x, embed_cls=False)["logits"][:, -1]
>     logits = repetition_penalty_processor(out, logits)
> ```

@vturrisi seems like moving all to HF is good, will do that including the penalty loss in new PR later today

vturrisi:
Great! Also move to  [MinLengthLogitsProcessor](https://huggingface.co/docs/transformers/v4.26.0/en/internal/generation_utils#transformers.MinLengthLogitsProcessor) for more clarity.

rom1504:
is this ready to merge ?

gpucce:
> is this ready to merge ?

@rom1504 yes will make new one for other small things

vturrisi:
@gpucce are you planning to add the penalty to the other PR?

gpucce:
> @gpucce are you planning to add the penalty to the other PR?

Yes I will add a logitprovessorlist with the penalty, min length, etc.

PR #405, opened 2023-02-03
add dual patchnorm as an option

lucidrains:
https://arxiv.org/abs/2302.01327v1

They have a section in there where they apply this to LiT

Seems like a free lunch, experiments are convincing :+1:

### DIFF
--- a/src/open_clip/model.py
+++ b/src/open_clip/model.py
@@ -30,7 +30,7 @@ class CLIPVisionCfg:
     image_size: Union[Tuple[int, int], int] = 224
     ls_init_value: Optional[float] = None  # layer scale initial value
     patch_dropout: float = 0.  # what fraction of patches to dropout during training (0 would mean disabled and no patches dropped) - 0.5 to 0.75 recommended in the paper for optimal results
-    dual_patchnorm: bool = False # whether to use dual patchnorm - would only apply the prelayernorm on each patch, as postlayernorm already exist in original clip vit design
+    input_patchnorm: bool = False # whether to use dual patchnorm - would only apply the input layernorm on each patch, as post-layernorm already exist in original clip vit design
     global_average_pool: bool = False  # whether to global average pool the last embedding layer, instead of using CLS token (https://arxiv.org/abs/2205.01580)
     attentional_pool: bool = False # whether to use attentional pooler in the last embedding layer
     n_queries: int = 256 # n_queries for attentional pooler
@@ -120,7 +120,7 @@ def _build_vision_tower(
             mlp_ratio=vision_cfg.mlp_ratio,
             ls_init_value=vision_cfg.ls_init_value,
             patch_dropout=vision_cfg.patch_dropout,
-            dual_patchnorm=vision_cfg.dual_patchnorm,
+            input_patchnorm=vision_cfg.input_patchnorm,
             global_average_pool=vision_cfg.global_average_pool,
             attentional_pool=vision_cfg.attentional_pool,
             n_queries=vision_cfg.n_queries,
--- a/src/open_clip/model.py
+++ b/src/open_clip/model.py
@@ -30,6 +30,7 @@ class CLIPVisionCfg:
     image_size: Union[Tuple[int, int], int] = 224
     ls_init_value: Optional[float] = None  # layer scale initial value
     patch_dropout: float = 0.  # what fraction of patches to dropout during training (0 would mean disabled and no patches dropped) - 0.5 to 0.75 recommended in the paper for optimal results
+    dual_patchnorm: bool = False # whether to use dual patchnorm - would only apply the prelayernorm on each patch, as postlayernorm already exist in original clip vit design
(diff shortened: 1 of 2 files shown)


lucidrains:
Screenshot from 2023-02-03 10-35-36

rwightman:
@lucidrains FWIW, I disagree with the post patch LN being before pos emb / class token, that feels like they didn't even bother to check properly / notice that CLIP ViT does it this way, I don't feel the extra code pass is worth it, but do feel having the option before the patch embed would be  nice....

rwightman:
as FlexiViT pointed out, normalizing after the patch embed helps with resizing the patches, I found that the CLIP style where it's after the pos embed too, definitely more robust to interpolating both patch and position embeddings without fine-tune than original vit, I don't think the proposed dualnorm would be so for pos embed...

lucidrains:
@rwightman sure thing, as long as we at least link to the paper and credit them for their experiments

they did show in the ablations that only post layernorm is worse than having both, but we should verify that on our end

rwightman:
> @rwightman sure thing, as long as we at least link to the paper and credit them for their experiments
> 
> they did show in the ablations that only post layernorm is worse than having both, but we should verify that on our end

Yeah, I found that a bit odd, all CLIP ViT has always been `patch, pos + token, ln, encoder` .. it seems at least as good to slightly better than original vit, so I really wonder about that ablation, and why they also didn't compare to after pos/token embed? Despite running some very compute expensive LiT w/ JFT-4B pretrain, they did not actually ablate against a CLIP ViT just the original. Makes me a bit suspect of the paper as a whole, guess I should try adding the pre (input ln) for a run to see what happens...

lucidrains:
yup, checkout this branch and do two small CLIP runs would be most definitive

MechCoder:
Someone pointed me to this PR. Thanks for trying it out so quickly and I felt the need to step-in to clarify.

> it seems at least as good to slightly better than original vit, so I really wonder about that ablation.

So the negative numbers in the table, are "ablations of Dual PatchNorm", that is they are the losses in accuracy from the Dual PatchNorm and not the losses in accuracy from the baseline. The other variants do improve over the baseline without Dual PatchNorm, i.e the original ViT. I will make this clearer in a revision.

> Despite running some very compute expensive LiT w/ JFT-4B pretrain, they did not actually ablate against a CLIP ViT just the original.

As you suggested, I tried this out (reply on twitter posted before you posted this comment) and performs on par with the Post ablation on imagenet ik. Moreover internally, LiT and Clip share the same codebase, and that does not have the norm after the positional embedding, see:
1 https://github.com/google-research/big_vision/blob/main/big_vision/models/proj/image_text/two_towers.py#L63
2. https://github.com/google-research/big_vision/blob/main/big_vision/models/vit.py#L177.

Since LiT has higher zero-shot numbers than Clip, I just retrained some LiT models.

> Makes me a bit suspect of the paper as a whole.

Tbh, this is adversarial and assumes bad faith without any good reason. If you want any specific clarifications from the paper, let me know.

rwightman:
@MechCoder don't think you understood my point there, CLIP (original) and LiT do not share the same codbase, LiT is using vanilla vit without an extra norm, CLIP has the extra norm after the pos embed, so when I say did not compare, I mean run equivalent LiT style training (obv the zero-shot for LiT training is higher since it pretrains on JFT) with the CLIP ViT (modified from original ViT)

https://github.com/openai/CLIP/blob/3702849800aa56e2223035bccd1c6ef91c704ca8/clip/model.py#L224-L230

lucidrains:
let's just run the experiments and see.. no big deal

lucidrains:
differences noted, we'll see if it can generalize with these little variations

mitchellnw:
I tried it out for a very short run. 20k iterations with batch size 65k with B/32. Accuracy improved by over 2% (52 -> 54%).

rwightman:
@mitchellnw great, did you notice any throughput differences? 

I think we can merge shortly, but do feel it's worth noting the second LN position is different than DualPatchNorm paper. This was part of why I was hesitant to call it that, it also leads one to believe if it's set to false it will involve no extra norm vs vanilla ViT (dual is in the name) since not everyone is aware of CLIP ViT being different, I thought input_norm makes it more clear what's being enabled/disabled....

In any case I fully agree with mentioning the dual paper / arxiv etc re the input norm.

mitchellnw:
yea sounds great to me

i didn't notice any throughput differences but that also may be because i'm using B/32

also updated comment above with more info on hparams

rwightman:
@mitchellnw B/32 should be worse case re noticing any slowdown, the extra LN is operating directly on the input and the rest of the net is less overhead than most of the other models we usually run. So, as long as it wasn't dataloading bound, should be good

lucidrains:
@MechCoder thank you for your paper! short and sweet! hope this plays out to another SOTA for the open source community :smile:

lucidrains:
@mitchellnw thank you for running the experiments :100:

lucidrains:
@rwightman i've changed the name as you suggested to `input_patchnorm`

rwightman:
@lucidrains tx!

PR #419, opened 2023-02-12
fix accum-grad training

### LINKED ISSUE #418: CoCa with accum-freq
Thanks for integrating CoCa!

When I tried coca with --accum-freq 2, I got this error:

```
  File "open_clip/src/training/main.py", line 426, in main
    train_one_epoch(model, data, loss, epoch, optimizer, scaler, scheduler, dist_model, args, tb_writer=writer)
  File "open_clip/src/training/train.py", line 141, in train_one_epoch
    model_out = model(images, texts, output_dict=True)
  File "... python3.9/site-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
TypeError: forward() got an unexpected keyword argument 'output_dict'
```

Looks like an easy fix by adding output_dict arg to CoCa models, but do you think if there will be other issues? @iejMac @rom1504

gpucce:
Should fix #418  

Hi @iejMac @rom1504 @usuyama, It should be the other way around, `output_dict` should be removed because it is passed to `create_model_and_transforms` in main.py, however there was also another small issue while accumulating grads, that should be fixed here.

I tried it and it seems to work ok, however, @iejMac could you maybe try it too and see if training goes as it should?

### DIFF
--- a/src/training/train.py
+++ b/src/training/train.py
@@ -137,12 +137,14 @@ def train_one_epoch(model, data, loss, epoch, optimizer, scaler, scheduler, dist
                 images = accum_images[j]
                 texts = accum_texts[j]
                 with autocast():
-                    model_out = model(images, texts, output_dict=True)
+                    model_out = model(images, texts)
                     logit_scale = model_out.pop("logit_scale")
-                    for key, val in accum_features:
+                    inputs = {}
+                    for key, val in accum_features.items():
                         accumulated = accum_features[key]
-                        accumulated = accumulated[:j] +  [model_out[key]] + accumulated[j + 1:]
-                    losses = loss(**accumulated, logit_scale=logit_scale, output_dict=True)
+                        inputs[key] = torch.cat(accumulated[:j] +  [model_out[key]] + accumulated[j + 1:])
+                    losses = loss(**inputs, logit_scale=logit_scale, output_dict=True)
+                    del inputs
                     total_loss = sum(losses.values())
                     losses["loss"] = total_loss
                 backward(total_loss, scaler)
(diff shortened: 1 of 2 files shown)


rwightman:
LGTM @gabrielilharco @mitchellnw do either of you have a quick check for grad accum? I've never used it so don't have an eyeball ref to tell if all is working on short order

mitchellnw:
also LGTM, nice work @gpucce. Able to do a small test on one GPU with coca training with  a small batch size just to make sure halving the batch size with --accum-freq 2 matches the curve without any grad accum?

gpucce:
Sure I'm running it, give me a moment and I tell you

gpucce:
> also LGTM, nice work @gpucce. Able to do a small test on one GPU with coca training with a small batch size just to make sure halving the batch size with --accum-freq 2 matches the curve without any grad accum?

They look the same

https://wandb.ai/gpucce/open-clip/reports/Untitled-Report--VmlldzozNTQwNzk4

mitchellnw:
great, any reason why contrastive loss is 0? would be good to know if things still work with contrastive loss with this if possible. otherwise looks good to convert out of draft and i'll merge

gpucce:
> great, any reason why contrastive loss is 0? would be good to know if things still work with contrastive loss with this if possible. otherwise looks good to convert out of draft and i'll merge

ah, that is the script for fine-tuning, running with contrastive and sharing in a bit

gpucce:
This has non zero contrastive too

https://api.wandb.ai/links/gpucce/0i1soih1

mitchellnw:
Thanks for testing! Will merge then.

usuyama:
Looking great! Thank you for fixing this very quickly! @gpucce

PR #424, opened 2023-02-13
vqa fine tune script

andrewwangva:


andrewwangva:
Don't worry about the argparse for now, because I prototyped on colab most of the time I passed an empty argument and just editited the default values of the args. I'll change it back soon when I wrap everything up.

wujianP:
Hi, thanks for your great script!
can you provide a demo script with hyperparameters that works?
thanks

PR #430, opened 2023-02-15
Add video support

iejMac:


iejMac:
Hmmm I don't see how we can reliably train this though. Let's say for L/14 max local batch size on a 40GB GPU is like a few hundred images so if a video has 100 frames at 1 FPS that gives us a local batch size <10 for rather short videos which probably still won't work very well unless we have more GPU's than we have

rom1504:
Gradient accumulation maybe ?

lucidrains:
@iejMac nice! i can contribute to this

i believe for video, we can do much more aggressive patch dropout in the beginning

lucidrains:
7baaai

iejMac:
@lucidrains cool! I'll start filling out the code a bit today. And yeah good idea with aggressive patch dropout. So currently we have:
- aggressive patch dropout
- grad accumulation

as some tricks to make this a bit more tractable. Anything to maximize batch size here will be really important.

iejMac:
ah also another thing I want to add here: simultaneous (image, video) - text training. I.e. the final model should be able to both handle temporal sequences and static images. The thing I'm unsure about is if we apply the image loss after the temporal decoder essentially saying "the time transformer should understand singular images" or before, just at the spatial transformer side

iejMac:
@lucidrains, does anything look wrong to you in the modeling code? Specifically [this](https://github.com/iejMac/open_clip/blob/80f41a068cf5c2db8b30ebf682a6ec188aea516b/src/open_clip/video_model.py#L50) ViViT class. I'm getting [very strange loss curves](https://wandb.ai/iejmac/open-clip/runs/2023_02_20-13_39_05-model_ViViT-B-32-lr_0.001-b_64-j_12-p_amp_bfloat16?workspace=user-iejmac) and was wondering if you might have an idea.

iejMac:
could also be the data loader code but I figured I'd ask you about the model since I'm comparing with [your vivit](https://github.com/lucidrains/vit-pytorch/blob/main/vit_pytorch/vivit.py)

lucidrains:
@iejMac nice! i'll do a code review later this week when i find some downtime

iejMac:
Next task - initialize spatial and text transformer from CLIP model

PR #437, opened 2023-02-19
fix slurm example

rom1504:


mitchellnw:
since @gabrielilharco and I both don't use those flags I don't think they're necessary / don't know why they're necessary for you

mitchellnw:
to get optimal EFA choices are
1. use these flags (works for nightly PT builds)
2. use aws pytorch install such as
```
conda install pytorch=1.12.1=aws* cudatoolkit=11.6 torchvision torchaudio \
--override-channels \
-c https://aws-pytorch.s3.us-west-2.amazonaws.com \
-c pytorch \
-c nvidia \
-c conda-forge
```
could be worth keeping around a minimal script for option 2?

Issue #246, opened 2022-11-23
mt5-base-ViT-B-32: Couldn't instantiate tokenizer


lopho:
This also fails the training & inference test for 'mt5-base-ViT-B-32'

```py
open_clip.get_tokenizer('mt5-base-ViT-B-32')
```
without sentencepiece
```
  File "/REDACTED/open_clip/src/open_clip/factory.py", line 75, in get_tokenizer
    tokenizer = HFTokenizer(config['text_cfg']['hf_tokenizer_name']) if 'hf_tokenizer_name' in config['text_cfg'] else tokenize
  File "/REDACTED/open_clip/src/open_clip/tokenizer.py", line 192, in __init__
    self.tokenizer = AutoTokenizer.from_pretrained(tokenizer_name)
  File "/REDACTED/open_clip/.venv/lib/python3.9/site-packages/transformers/models/auto/tokenization_auto.py", line 619, in from_pretrained
    return tokenizer_class.from_pretrained(pretrained_model_name_or_path, *inputs, **kwargs)
  File "/REDACTED/open_clip/.venv/lib/python3.9/site-packages/transformers/tokenization_utils_base.py", line 1777, in from_pretrained
    return cls._from_pretrained(
  File "/REDACTED/open_clip/.venv/lib/python3.9/site-packages/transformers/tokenization_utils_base.py", line 1932, in _from_pretrained
    tokenizer = cls(*init_inputs, **init_kwargs)
  File "/REDACTED/open_clip/.venv/lib/python3.9/site-packages/transformers/models/t5/tokenization_t5_fast.py", line 134, in __init__
    super().__init__(
  File "/REDACTED/open_clip/.venv/lib/python3.9/site-packages/transformers/tokenization_utils_fast.py", line 120, in __init__
    raise ValueError(
ValueError: Couldn't instantiate the backend tokenizer from one of: 
(1) a `tokenizers` library serialization file, 
(2) a slow tokenizer instance to convert or 
(3) an equivalent slow tokenizer class to instantiate and convert. 
You need to have sentencepiece installed to convert a slow tokenizer to a fast one.
```
with sentencepiece & latest protobuf (4.21.9)
```
  File "/REDACTED/open_clip/src/open_clip/factory.py", line 75, in get_tokenizer
    tokenizer = HFTokenizer(config['text_cfg']['hf_tokenizer_name']) if 'hf_tokenizer_name' in config['text_cfg'] else tokenize
[...]

PR #449, opened 2023-02-24
update protobuf minor version

mldangelo:
This allows protobuf to be any version < 4.  This should help users of poetry that require both tensorflow 2.11 and open-clip-torch.


rwightman:
@lopho @rom1504 re #246 ... what's the max version of protobuf we can support?  will < 4 cause any issues with mt5? I think I'd rather break one model (ie people that wan't mt5 have to install a more restrictive protobuf) than break for everyone that needs other libs in their env.. hmm

dsully:
+1 to this change. I'm seeing this right now due to the (effective) pin:
```
Unable to find a resolution for protobuf
because of the following conflicts:
  protobuf<3.20,>=3.9.2 (from tensorflow@2.11.0)
  protobuf<4,>=3.9.2 (from tensorboard@2.11.0)
  protobuf==3.20.* (from open-clip-torch@2.14.0)
```

rwightman:
so 3.9.2 would be need for both tf and tb to work, so I guess the q is does sentencepiece work with that, we'll have to break one or the other if there isn't a common version for all cases ... the joys of protobuf :/

rwightman:
Since we dont have any petrained weights for mt5, going to merge and deal with any issues related to sentencepiece later, it should be an optional import and not a req dep anyways if there are no weights yet...

PR #460, opened 2023-03-08
Remove pytorch MHA from attentionl pooler

gpucce:
@rwightman this is an attempt at changing the attentional pooler so that things are more flexible, (plus torchscript might work) and ideally in the future it might also be used to cache key_values in generation if applied inside the transformers @iejMac, @rom1504.

The idea is eesentially to reimplement how it is done in pytorch, with less options. Working toward matching pytorch exactly for now


gpucce:
@rwightman @rom1504 @iejMac this is the first step towards caching past key values and it is now working, however I would like to add tests and I am not sure how. 

The reason is that since I implement the attn by hand it is hard to make the regression test work with a not pretrained mode because the initialization ends up different, not sure how I can fix that. Otherwise I can do it by testing on a pretrained model, however all existing regression tests are done differently. I tested locally and with the pretrained ones it works. 

How would you prefer I do this?

P.S. Sorry this is taking a lot of time but I can never find time to work on it

rwightman:
FWIW, running black formatting, etc is going to significanlty expand the deltas and increase probability of merge conflicts for other PRs

gpucce:
> FWIW, running black formatting, etc is going to significanlty expand the deltas and increase probability of merge conflicts for other PRs

Ooh sorry, won't have it when the PR is ready.

gpucce:
@iejMac Too much work for liitle gain, to address the issue this was meant for I opened #559

PR #470, opened 2023-03-20
Fix pytorch versison

quintant:
Need to add an upper bound to the Pytorch version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-torch>=1.9.0
+torch>=1.9.0,<2.0.0
 torchvision
 regex
 ftfy


rwightman:
@quintant why? I admit I haven't tried the official 2.0 release yet, but I've trained some significant models over past few months with 2.0 nightlies without issues (they were actually needed to train convnext models)

quintant:
> @quintant why? I admit I haven't tried the official 2.0 release yet, but I've trained some significant models over past few months with 2.0 nightlies without issues (they were actually needed to train convnext models)

I got some strange issues where the tensors were not on the same device even though both were on cpu, and that got fixed when I reverted the PyTorch version.

rwightman:
@quintant okay, that may be an issue to figure out but not a reason to change requirements, so will close

PR #482, opened 2023-04-10
--train-num-samples takes priority over guessing the dataset size

gabrielilharco:
Currently `--train-num-samples` is not used if `get_dataset_size` succeeds in finding the dataset size. This is not ideal because if metadata such as a `sizes.json` file exists, we currently can't control the virtual dataset size. This means we can't have more frequent checkpoints / evaluate more often in case the dataset is big, which is undesirable.

This PR addresses this by giving priority to the `--train-num-samples` flag, only trying to read the metadata files if the flag is not used.

### DIFF
--- a/src/training/data.py
+++ b/src/training/data.py
@@ -328,7 +328,7 @@ def get_wds_dataset(args, preprocess_img, is_train, epoch=0, floor=False, tokeni
     assert input_shards is not None
     resampled = getattr(args, 'dataset_resampled', False) and is_train
 
-    num_shards = len(expand_urls(input_shards))
+    num_shards = len(expand_urls(input_shards)[0])
 
     if is_train:
         if args.train_num_samples is not None:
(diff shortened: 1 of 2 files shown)


rwightman:
@gabrielilharco good to fix yes, w/ current solution it may expand same shard list 3x though which could be a bit excessive for very lage datasets

for validation num_shards can be set to 0
for train could keep the old approach where num samples method still returns the # shards since it's already calculated it when --train-num-samples not set, and then a manual expand call if --train-num-samples set so expand is done at most 2x (done again in ResampledShard...)

gabrielilharco:
Thanks @rwightman! I pushed an update, let me know if this is what you had in mind

rwightman:
@gabrielilharco I pulled this PR into another one since I couldn't push to yours for some reason, added two commits on the end. Merged

PR #488, opened 2023-04-14
int8 support

mitchellnw:
This PR introduces beta support for int8 training and inference. You can enable int8 training with `--use-bnb-linear SwitchBackLinearGlobal` or `--use-bnb-linear SwitchBackLinearGlobalMemEfficient`. For CLIP VIT-Huge this should currently correspond to a 10% training speedup with negligable accuracy difference. 

More speedups coming when the attention layer is refactored so that linear layers may be replaced there, too. However, that will require more changes so best to get this merged first.

### DIFF
--- a/src/training/main.py
+++ b/src/training/main.py
@@ -250,6 +250,7 @@ def main(args):
         print(f'=> replacing linear layers with {args.use_bnb_linear}')
         linear_replacement_cls = getattr(bnb.nn.triton_based_modules, args.use_bnb_linear)
         replace_linear(model, linear_replacement_cls)
+        model = model.to(device)
 
     random_seed(args.seed, args.rank)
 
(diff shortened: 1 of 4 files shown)


rwightman:
@mitchellnw looks low risk, ready to merge?

mitchellnw:
@rwightman yep, ready to merge!

src/training/main.py
```diff
@@ -277,6 +287,8 @@ def main(args):
         if args.ddp_static_graph:
             # this doesn't exist in older PyTorch, arg only added if enabled
             ddp_args['static_graph'] = True
+        if args.use_bnb_linear is not None:
```
rwightman, line 290: couldn't this be put right after replace_linear() in the previous `if use_bnb_linear` block?

rwightman: @mitchellnw one q here

mitchellnw: yep good point that simplifies -- i'll make the change and run a quick test to make sure all good.

mitchellnw: fixed, thanks @rwightman!

PR #500, opened 2023-04-18
Add EVA models (via timm backbone), torch.compile support, more

rwightman:
* Add EVA and EVA02 CLIP models via timm backbone w/ the support for swiglu, extra norm layers, post norm, ROPE embed (seemed cleaner than adding a whole nother model class here since it's aleady `timm` style
* Add support for torch.compile, need to call model via forward() for easy use so made forward args optional to allow
* Add 'pure' bf16 and fp16 precision modes as the current ones were inherited from OpenAI CLIP and are actually a form of manual mixed-precision. This is useful for the 'E' EVA model that won't fit on a 24GB card for inference without
* Tweak HF hub tags and upload push script
* Remove OpenAI native jit loading, always convert state dict to non-jit model before re-jitting if enabled. OpenAI jit models having issues loading in PyTorch 2.0 for the trace / patching.


cachett-ML:
Can you explain how one should proceed to use EVA models using open_clip?
Thanks you!!

PR #504, opened 2023-04-22
Add text-text (audio) CLIP

marianna13:
Text-Text (audio) CLIP (based on [#323](https://github.com/mlfoundations/open_clip/pull/323) but one of the text tokens are actually audio encoded with [Encodec](https://github.com/facebookresearch/encodec) (can add other encoding methods). One of the models is just BERT (for real text) and the other one is [LSG BERT](https://huggingface.co/ccdv/lsg-bert-base-uncased-4096).


rwightman:
Oyo, so, where does this sit relative to #323? Is the intent to have that one cleaned up and merged and then this one will merge those changes before being merged?

marianna13:
Hi @rwightman ! I added description, basically we pretend that encoded audio is just text tokens and treat it respectively. I trained it on a small subset of LAION630k (basically [VGGSound](https://github.com/marianna13/audio-dataset/blob/marianna/data_cards/VGGSound/VGGSound.md)) for just a little and it gave me not so bad results. I'm just wondering is it even worth it to have such a model and what other people think of it :)

PR #507, opened 2023-04-25
Enabling SMP

rahul003:
Below changes show how to integrate SMP into the OpenClip training script. Currently it only does changes for training,  evaluation needs similar changes. 

Notably the following main changes
- Init SMP similar to what is done for Horovod in this script.
- Wrappers for model and optimizer.
- The style of mixed precision is different with SMP. Instead of using amp/autocast/gradScaler, we have our own mixed precision within SMP which can be activated by setting flag for dtype when initializing SMP. The layernormFp32 in this repo is not compatible with that as the weight would be in half precision in SMP. Changing to SMP's LayerNorm can address this issue.
- There are couple of instances in the model, where a module works on params that were not created in its own module. With Sharded data parallelism in SMP we need to register these parameters so they are gathered before execution.
- Some API changes for checkpointing, so we can handle the sharded states across GPUs.
- Training script needs to be encapsulated within smp.step annotated function and model.backward needs to be called for backward.

### DIFF
--- a/src/training/data.py
+++ b/src/training/data.py
@@ -5,11 +5,10 @@ import math
 import os
 import random
 import sys
-import time
+import braceexpand
 from dataclasses import dataclass
 from multiprocessing import Value
 
-import braceexpand
 import numpy as np
 import pandas as pd
 import torch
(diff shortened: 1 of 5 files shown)


rahul003:
Currently it doesn’t use sync-batch-norm but if I believe it can work with SMP too by passing process_group=smp.get_process_group()

PR #517, opened 2023-05-02
Add DataComp models

gabrielilharco:


mitchellnw:
why add all the models? this makes it a bit confusing -- what about just the best one?

ludwigschmidt:
I don't think there is anything wrong with releasing many models, and the models may be useful for others.

Having said that, I agree that a huge README can be confusing. What about we move most of the models (not just DataComp but generally) into a separate file "model zoo" (or something similar) and then highlight only the most important models in the main README while linking to the model zoo for the full list of all models?

As a side note, we could also structure the list of DataComp models a bit better by adding sub-headings for the different scales.

More broadly, refactoring the README into separate files could be a good idea?

mitchellnw:
ya i think the concern is that openclip is getting a bit confusing and too much going on -- there are lots of large PRs -- so if we can do small PRs / small changes thats probably good

gabrielilharco:
I moved all large, medium and small scale models to a separate readme and added a pointer on the main readme, where I also kept the xlarge models

mitchellnw:
nice!

PR #523, opened 2023-05-09
Implement Locking of Text Tower for `CLIP` Models

rsomani95:
Currently, the `lock-text.*` args only work when using a `HFTextEncoder`. This PR adds a `lock_text_tower` method to the 'native' `CLIP` class as well.

There's two aspects that are different from the `HFTextEncoder` that I'd love feedback on:
1. `embeddings` in this case are two separate layers, whereas HF has just one
2. `self.text_projection` and `self.ln_final` are passed in as the last two layers respectively. The HF class doesn't have an analogue. So, to unfreeze the last transformer block, you need to pass in `lock-text-unlocked-layers 3`, which feels awkward.

### DIFF
--- a/src/open_clip/model.py
+++ b/src/open_clip/model.py
@@ -205,6 +205,37 @@ class CLIP(nn.Module):
         # lock image tower as per LiT - https://arxiv.org/abs/2111.07991
         self.visual.lock(unlocked_groups=unlocked_groups, freeze_bn_stats=freeze_bn_stats)
 
+    def lock_text_tower(self, unlocked_layers: int = 0, freeze_layer_norm: bool = True):
+        # Unlike the `self.visual` component, the text encoder is embedded into the main class
+        # So we need to do all the work here itself
+        # See `self.encode_text()` for order of ops. We want to freeze the last layers first
+        embeddings = [self.token_embedding, self.positional_embedding]
+        resblocks = self.transformer.resblocks
+        final_text_projection = self.text_projection
+        final_ln = self.ln_final
+
+        # TODO: Make both embeddings behave as one module, and the final_ln + final_text_projection
+        # as a single module as well? Not sure how the final layers are structured in HF
+        modules = [*embeddings, *resblocks, final_text_projection, final_ln]
+
+        if (not unlocked_layers) or (unlocked_layers == 0):  # full freezing
+            modules_to_lock = modules
+        else:
+            modules_to_lock = modules[:-unlocked_layers]
+
+        for module in modules_to_lock:
+            # `self.text_projection` and `self.positional_embedding`
+            if isinstance(module, nn.Parameter):
+                module.requires_grad = False
+
+            # All other modules
+            elif isinstance(module, nn.Module):
+                for n, p in module.named_parameters():
+                    p.requires_grad = (not freeze_layer_norm) if "LayerNorm" in n.split(".") else False
+
+            else:
+                raise TypeError(f"Encountered unexpected module type {type(module)} for module {module}")
+
     @torch.jit.ignore
     def set_grad_checkpointing(self, enable=True):
         self.visual.set_grad_checkpointing(enable)


rwightman:
@rsomani95 thanks for the contribution

For this to work in all current cases, the bulk of the logic needs to be in a free function because this should work with both CLIP and CustomTextCLIP (w/ builtin text-tower) via the TextTransformer module, So the TextTransformer needs a lock method that calls the free fn, and the CLIP needs the lock_text_tower method you have that also calls the free fn, passing both passing in the list of modules/parameters... 

Looking at the vision lock as  a template, the grouping strategy used there is a solultion to your awkwardness and we should be consistent with that, we grouped the input + abs pos embed + any pre-ln into the first group, each transformer block as another group, lumped the last transformer block with the final LN, and final projection as its own...

```
        if unlocked_groups != 0:
            groups = [
                [
                    self.conv1,
                    self.class_embedding,
                    self.positional_embedding,
                    self.ln_pre,
                ],
                *self.transformer.resblocks[:-1],
                [
                    self.transformer.resblocks[-1],
                    self.ln_post,
                ],
                self.proj,
            ]

            def _unlock(x):
                if isinstance(x, Sequence):
                    for g in x:
                        _unlock(g)
                else:
                    if isinstance(x, torch.nn.Parameter):
                        x.requires_grad = True
                    else:
                        for p in x.parameters():
                            p.requires_grad = True
```

rsomani95:
@rwightman that freezing code is a lot cleaner, I've updated mine to mimic the vision lock.

Also implemented the free function as I understood it. Let me know if this is in line with what you were describing above.

dfan:
The `lock_text_transformer` function is not correct because it doesn't add a case for handling a list of modules. It will raise a TypeError

rsomani95:
@dfan with which model do you encounter that error?

Yangr116:
The ```token_embedding``` is ```nn.Embedding```,  so it will cause a type error.

rohun-tripathi:
@rsomani95 With ViT-H-14-378-quickgelu and ViT-H-14-336 I see a type error.
Which models did you test?

I temporarily resolved it using the following version but am yet to run a complete training - 

    def lock_text_transformer(transformer: TextTransformer, unlocked_layers: int = 0, freeze_layer_norm: bool = True):

        groups = [
            [transformer.token_embedding, transformer.positional_embedding],
            *transformer.transformer.resblocks[:-1],
            [transformer.transformer.resblocks[ -1], transformer.ln_final],
            transformer.text_projection,
        ]

        def _freeze(modules, freeze_layer_norm: bool = True):
            for module in modules:
                # `CLIP.text_projection` and `CLIP.positional_embedding`
                if isinstance(module, nn.Parameter):
                    print(f"Freezing module {module} which is a nn.Parameter")
                    module.requires_grad = False

                # All other modules
                elif isinstance(module, nn.Module):
                    print(f"Freezing module {module} which is a nn.Module")
                    for n, p in module.named_parameters():
                        p.requires_grad = (not freeze_layer_norm) if "LayerNorm" in n.split(".") else False

                elif isinstance(module, list):
                    for sub_module in module:
                        
                        if isinstance(sub_module, nn.Parameter):
                            print(f"Freezing sub_module {sub_module} which is a nn.Parameter")
                            sub_module.requires_grad = False
                        
                        elif isinstance(sub_module, nn.Module):
                            print(f"Freezing sub_module {sub_module} which is a nn.Module")
                            for n, p in sub_module.named_parameters():
                                p.requires_grad = (not freeze_layer_norm) if "LayerNorm" in n.split(".") else False

                        else:
                            raise TypeError(f"Encountered unexpected module type {type(sub_module)} for sub_module {sub_module}")
                else:
                    raise TypeError(f"Encountered unexpected module type {type(module)} for module {module}")

rsomani95:
I think this should be good to go now.

@rwightman I couldn't think of a more robust way to extract the LayerNorm layers inside the resblocks:
https://github.com/mlfoundations/open_clip/blob/f065ee612d43ee654a35814b6c37ffdd89fac27c/src/open_clip/transformer.py#L830-L839

The assumption here is that the only norm layer being used is LayerNorm. I think, but am not 100% sure, that all existing text models are using LayerNorm?

rsomani95:
@rwightman I couldn't think of a more robust way to extract the LayerNorm layers inside the resblocks:
https://github.com/mlfoundations/open_clip/blob/f065ee612d43ee654a35814b6c37ffdd89fac27c/src/open_clip/transformer.py#L830-L839

The assumption here is that the only norm layer being used is LayerNorm. I think, but am not 100% sure, that all existing text models are using LayerNorm?

alexisdrakopoulos:
It would still be nice to get this!

alexisdrakopoulos:
It seems to work for ViT-B-16-SigLIP2

PR #530, opened 2023-05-12
Fix code not working with Python 3.7

bryant1410:
`Literal` is introduced in Python 3.8. For it to work on previous Python versions, there's `typing_extensions`, which is likely already transitively installed by the direct dependencies.

### DIFF
--- a/requirements-training.txt
+++ b/requirements-training.txt
@@ -10,4 +10,3 @@ huggingface_hub
 transformers
 timm
 fsspec
-typing_extensions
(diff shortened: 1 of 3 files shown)


bryant1410:
Another option is to stop supporting Python 3.7, which reaches EoL soon: https://devguide.python.org/versions/

rwightman:
hrmm, damn 3.7, for now would be better to just remove the use of Literal and change to str....

Issue #549, opened 2023-06-24
build_cls_mask() in CoCa TextTransfotmer 


yiren-jian:
TL, DR: current implementation of `build_cls_mask()` produces `cls_mask` for [CLS] being as the first token. But in CoCa, [CLS] is the end token.

In [Issue 312](https://github.com/mlfoundations/open_clip/pull/312), `build_cls_mask()` was introduced by @gpucce in `TextTransformer` in CoCa to  "preventing the CLS token at the end of the sequence from attending to padded tokens".

```python
# https://github.com/mlfoundations/open_clip/blob/main/src/open_clip/transformer.py#L587
def build_cls_mask(self, text, cast_dtype: torch.dtype):
        cls_mask = (text != self.pad_id).unsqueeze(1)
        cls_mask = F.pad(cls_mask, (1, 0, cls_mask.shape[2], 0), value=1.0)
        additive_mask = torch.empty(cls_mask.shape, dtype=cast_dtype, device=cls_mask.device)
        additive_mask.fill_(0)
        additive_mask.masked_fill_(~cls_mask, float("-inf"))
        additive_mask = torch.repeat_interleave(additive_mask, self.heads, 0)
        return additive_mask
```

Taking `text = torch.tensor([[1,2,3,4,0,0,0]])` as an example,

```python
import torch
import torch.nn.functional as F

text = torch.tensor([[1,2,3,4,0,0,0]])  ### batch size 1, sequence 4 with 3 padding (pad_id=0)

pad_id = 0
cls_mask = (text != pad_id).unsqueeze(1)
cls_mask = F.pad(cls_mask, (1, 0, cls_mask.shape[2], 0), value=1.0)
additive_mask = torch.empty(cls_mask.shape)
additive_mask.fill_(0)
additive_mask.masked_fill_(~cls_mask, float("-inf"))
print(additive_mask)
```

This output 
```
tensor([[[0., 0., 0., 0., 0., 0., 0., 0.],
         [0., 0., 0., 0., 0., 0., 0., 0.],
         [0., 0., 0., 0., 0., 0., 0., 0.],
         [0., 0., 0., 0., 0., 0., 0., 0.],
         [0., 0., 0., 0., 0., 0., 0., 0.],
         [0., 0., 0., 0., 0., 0., 0., 0.],
         [0., 0., 0., 0., 0., 0., 0., 0.],
         [0., 0., 0., 0., 0., -inf, -inf, -inf]]])
```

[...]

PR #550, opened 2023-06-24
Mask CLS at the end of seq

gpucce:
@yiren-jiran thanks for finding the bug, in #549 it looks like I made a mistake in creating the cls mask.

This PR should be the fix, will have to look how much changing this affects models performance.

### DIFF
--- a/src/open_clip/transformer.py
+++ b/src/open_clip/transformer.py
@@ -499,7 +499,7 @@ class VisionTransformer(nn.Module):
 
         if self.output_tokens:
             return pooled, tokens
-        
+
         return pooled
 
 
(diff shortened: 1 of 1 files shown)


yiren-jian:
I believe fixing this issue requires re-training CoCa models. 

However, it's not supervised to me that the current trained models working well: [CLS] token was still able to "see" all the relevant preceding text, but not itself. 

Fixing the bug without re-training models would hurt retrieval/classification performance, I guess.

gpucce:
Indeed, likely needs retraining we this change.

gpucce:
@rwightman what would be the best way to address this? Does it make sense to retrain the model?

rom1504:
but first fix the CI here

gpucce:
@rom1504 it appears to be an issue with caching older test run times or smth similar, removing

https://github.com/mlfoundations/open_clip/blob/fb72f4db1b17133befd6c67c9cf32a533b85a321/.github/workflows/ci.yml#L96-L98

seems to fix it, however I am not super confident about github ci, 
if you or someone else who is can check it would probably help

gpucce:
@rom1504 @iejMac so I found that pluggy was not versioned, adding it to the `requirements-test.txt` file fixed the issue in the appropriate way I believe.

gpucce:
@rom1504 sorry for several tags, I think it was a caching issue but honestly not sure how it got fixed

gpucce:
Moved to #551

PR #553, opened 2023-06-27
Tiny CI fix

gpucce:


gpucce:
@rwightman @rom1504 this time I should have got it right.
If there aren´t cached files (in this case it was because of newer python 3.8 version) 
`.test_durations` was created with `touch` but that is not json readable and so `pytest` would throw an error.

.github/workflows/ci.yml
```diff
@@ -93,7 +93,12 @@ jobs:
     - name: Unit tests
       run: |
         source .env/bin/activate
-        touch .test_durations
+        if [[ -f .test_durations ]]
+        then
+          touch .test_durations
```
rom1504, line 98: if the file already exists touch does nothing; so let's just do the echo only if the file doesn't exist

gpucce:
pytest-split is fine with missing duration-files

https://github.com/jerry-git/pytest-split/blob/0798ea1dfd6aa6ae962799e32103003939e52681/src/pytest_split/plugin.py#L128C1-L132

so I flipped the logic because it seems better

gpucce:
@rom1504 if you have time could you review this? All new PRs fail CI without this I think

PR #559, opened 2023-06-28
split pooling

gpucce:
This should make attentional pooling identical to the original paper @iejMac


rwightman:
This was added in the SigLIP/CLIPA merge in a way that hopefully will maintain backward compat for existin gmodels while allowing exploration of either cascade or paralllel pooling described in paper. If any issues/updates, will do a new pr. Tx!

PR #561, opened 2023-06-29
default clip loss

gpucce:
to avoid computing clip loss when it is zero we still need a tensor otherwise it errors somewhere else.


iejMac:
lgtm, I encountered this earlier as well, comes up when you try to fine-tune captioner. cc: @rom1504

gpucce:
I think it still gives trouble with ddp when set to 0 because some parameters fall out of loss computation

PR #613, opened 2023-08-28
Log wandb step using wandb native step arg in addition to the "step" key.

gabrielilharco:
Currently we are passing `step` to wandb as another metric instead of using the `step` argument to `wandb.log` (see https://docs.wandb.ai/ref/python/log). This causes two "step" variables to be logged and can cause inconsistencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,10 +415,16 @@ When training a RN50 on YFCC the same hyperparameters as above are used, with th
 
 Note that to use another model, like `ViT-B/32` or `RN50x4` or `RN50x16` or `ViT-B/16`, specify with `--model RN50x4`.
 
-### Launch tensorboard:
+### Logging
+
+For tensorboard logging, run:
 ```bash
 tensorboard --logdir=logs/tensorboard/ --port=7777
 ```
+
+For wandb logging, we recommend looking at the `step` variable instead of `Step`, since the later was not properly set in earlier versions of this codebase.
+For older runs with models trained before https://github.com/mlfoundations/open_clip/pull/613, the `Step` variable should be ignored.
+For newer runs, after that PR, the two variables are the same.
 
 ## Evaluation / Zero-Shot
 
--- a/src/training/train.py
+++ b/src/training/train.py
@@ -225,7 +225,7 @@ def train_one_epoch(model, data, loss, epoch, optimizer, scaler, scheduler, dist
                     tb_writer.add_scalar(name, val, step)
                 if args.wandb:
                     assert wandb is not None, 'Please install wandb.'
-                    wandb.log({name: val, 'step': step})
+                    wandb.log({name: val}, step=step)
 
             # resetting batch / data time meters per log window
             batch_time_m.reset()
--- a/src/training/train.py
+++ b/src/training/train.py
@@ -219,14 +219,17 @@ def train_one_epoch(model, data, loss, epoch, optimizer, scaler, scheduler, dist
             }            
             log_data.update({name:val.val for name,val in losses_m.items()})
 
-            for name, val in log_data.items():
-                name = "train/" + name
-                if tb_writer is not None:
-                    tb_writer.add_scalar(name, val, step)
-                if args.wandb:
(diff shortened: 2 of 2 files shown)


rom1504:
Making this is change now would make it difficult to compare new and old runs

Is there any benefit?

gabrielilharco:
We can keep  `'step': step` inside the dict for backwards compatibility. Probably still good to fix the wandb step, I see no harm in doing so

EIFY:
Hmm, shouldn't we also fix the logging line in `evaluate()`?
https://github.com/mlfoundations/open_clip/blob/91b7b512d26899e8b660414daea509d5bda38aec/src/training/train.py#L341-L344
To be consistent we probably should also use the number of training steps so far here, e.g.
```python
if args.wandb:
    assert wandb is not None, 'Please install wandb.'
    dataloader = data['train'].dataloader
    num_batches_per_epoch = dataloader.num_batches // args.accum_freq
    step = num_batches_per_epoch * epoch
    for name, val in metrics.items():
        wandb.log({f"val/{name}": val, 'epoch': epoch}, step=step)
```

EIFY:
Upon closer look I think we can consolidate prefixing and `wandb.log()` calls to reduce remote calls over the network, e.g.:
```python
# In train_one_epoch():

            log_data = {"train/" + k: v for k, v in log_data.items()}
            if tb_writer is not None:
                for name, val in log_data.items():
                    tb_writer.add_scalar(name, val, step)
            log_data['step'] = step
            if args.wandb:
                assert wandb is not None, 'Please install wandb.'
                wandb.log(log_data, step=step)
```
```python
# In evaluate():

    log_data = {"val/" + k: v for k, v in metrics.items()}

    if args.save_logs:
        if tb_writer is not None:
            for name, val in log_data.items():
                tb_writer.add_scalar(name, val, epoch)

        with open(os.path.join(args.checkpoint_path, "results.jsonl"), "a+") as f:
            f.write(json.dumps(metrics))
            f.write("\n")

    if args.wandb:
        assert wandb is not None, 'Please install wandb.'
        dataloader = data['train'].dataloader
        num_batches_per_epoch = dataloader.num_batches // args.accum_freq
        step = prev_steps + num_batches_per_epoch * (epoch - prev_epochs)
        log_data['epoch'] = epoch
        wandb.log(log_data, step=step)
```
Overall I strongly vote for fixing this. Current `wandb.log({name: val, 'step': step})` call is confusing by itself and the issue is exacerbated by how we iterate over `log_data.items()` since by default `wandb`'s internal "step" increments after each call. Even worse, `len(log_data)` depends on the loss we use (e.g. `DistillClipLoss` adds an extra `"distill_loss"`), so even comparable training runs may not be aligned along the "step" x-axis on the `wandb` dashboard.

EIFY (approved):

gabrielilharco:
Thanks @EIFY. Just pushed some changes. Tested on a run and wandb is looking good. We still have a lowercase `step` variable being logged so everything is backwards compatible.

rom1504:
the current `Step` should simply not be used in wandb ui

I think changing this will make it really confusing for comparing old runs where Step is meaningless with new runs when Step is meaningful

if we do change it, can we at least add a big warning in some logging section in the readme?
and indeed also keep the old `step` key

gabrielilharco:
Even with the change, people can keep doing comparisons the exact same way they did before, by looking at lowercase `step`. 

I'm also fine with keeping things as they are. I think it's cleaner and less confusing to have a `Step` that's meaningful, but I also don't think it's a big deal. So up to you @rom1504

EIFY:
@rom1504 wandb ui defaults to its own `Step` and honestly I didn't know it's customizable until now. I think it would be nice for the dashboard to behave as expected at the first sight, and now there is the additional benefit of aligning the validation metrics by `Step`.

EIFY:
@rom1504 Could we merge this? I think concerns have been addressed.

rom1504 (approved):

rom1504:
Yes merged

Issue #618, opened 2023-09-04
Request for Sigmoid Loss Integration: SigLip


fabiozappo:
Hi there,

First of all, I'd like to express my appreciation for the outstanding work you've done on this project. Your efforts are truly commendable!

I recently had the opportunity to read the [SigLip](https://arxiv.org/pdf/2303.15343.pdf) research paper, and the authors of the paper propose that replacing the conventional softmax loss with a sigmoid loss could potentially enhance a model's learning capabilities, particularly when dealing with lower batch sizes. Given the potential benefits highlighted in the paper, I'm curious to know if there are any plans to integrate this approach into your repository.

rom1504:
@rwightman Could you describe what you tried on this ?

rwightman:
I had this 80% of the way but hadn't debugged the isend/irecv code to do the neighbour shifting (the paper authors use jax.lax.ppermute for this which is not available as pytorch primitive).

I'll see if I can get it into a state where it does something soonish... and then push a PR for others to look at, help polish/test, etc

rwightman:
@rom1504 @fabiozappo what I have is here, I tried working on the distributed part a bit more, but not sure I've got it behaving properly... at least it seems to not converge, or converge very poorly vs non-distributed ... currently trying a new autograd.Function approach which exchanges grad in opposite dir...

PR #634, opened 2023-09-15
SigLIP impl

rwightman:
Re #618

### DIFF
--- a/src/open_clip/factory.py
+++ b/src/open_clip/factory.py
@@ -13,7 +13,7 @@ from .constants import OPENAI_DATASET_MEAN, OPENAI_DATASET_STD
 from .model import CLIP, CustomTextCLIP, convert_weights_to_lp, convert_to_custom_text_state_dict,\
     resize_pos_embed, get_cast_dtype
 from .coca_model import CoCa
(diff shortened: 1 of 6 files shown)


lucasb-eyer:
Can't comment on the distributed part of the code as I don't know that part of PyTorch, but the rest (loss details, bias/temp/inits) LGTM.

rwightman:
@lucasb-eyer thanks for taking a look, yeah the dist part is where a lot of the risk is, but seems to be behaving on local cc12m runs comparing single to 4x GPU.

lucasb-eyer:
FYI: in our code, Basil implemented a small unit-test checking both formulations for "almost equalness" of chunked vs non-chunked, this gave us good reassurance in the implementation (+looking at profiler for memory use).

rwightman:
I've tested
* `convnext_tiny` on cc12m old InfoNCE run vs new SigLIP (36.13 vs 36.46 zero-shot in1k)
* initial convergence w/ siglip + grad accum enabled
* initial convergence of custom text and original clip models w/o siglip (original CLIP InfoNCE loss)
* initial convergence with bidirection exchange and unidirectional
* validating several existing models

Will merge shortly to prevent this getting stale

Issue #632, opened 2023-09-14
Error when using train-data-upsampling-factors


humzaiqbal:
Hi! I'm trying to use multiple datasets when fine tuning OpenCLIP and I want to use train-data-upsampling-factors while doing so however I ran into an issue when trying to get `data["val"]`
```
AssertionError: --train_data_upsampling_factors is only supported when sampling with replacement (with --dataset-resampled).
    data["val"] = get_dataset_fn(args.val_data, args.dataset_type)(
  File "/home/ubuntu/research_nfs/humza/open_clip/src/training/data.py", line 357, in get_wds_dataset
    assert args.train_data_upsampling_factors is None,\
AssertionError: --train_data_upsampling_factors is only supported when sampling with replacement (with --dataset-resampled).
    assert args.train_data_upsampling_factors is None,\
AssertionError: --train_data_upsampling_factors is only supported when sampling with replacement (with --dataset-resampled).
ERROR:torch.distributed.elastic.multiprocessing.api:failed (exitcode: 1) local_rank: 0 (pid: 9027) of binary: /home/ubuntu/miniconda3/envs/rapids-22.06/bin/python3.9
Traceback (most recent call last):
  File "/home/ubuntu/miniconda3/envs/rapids-22.06/bin/torchrun", line 8, in <module>
    sys.exit(main())
  File "/home/ubuntu/miniconda3/envs/rapids-22.06/lib/python3.9/site-packages/torch/distributed/elastic/multiprocessing/errors/__init__.py", line 345, in wrapper
    return f(*args, **kwargs)
  File "/home/ubuntu/miniconda3/envs/rapids-22.06/lib/python3.9/site-packages/torch/distributed/run.py", line 724, in main
    run(args)
  File "/home/ubuntu/miniconda3/envs/rapids-22.06/lib/python3.9/site-packages/torch/distributed/run.py", line 715, in run
    elastic_launch(
  File "/home/ubuntu/miniconda3/envs/rapids-22.06/lib/python3.9/site-packages/torch/distributed/launcher/api.py", line 131, in __call__
    return launch_agent(self._config, self._entrypoint, list(args))
  File "/home/ubuntu/miniconda3/envs/rapids-22.06/lib/python3.9/site-packages/torch/distributed/launcher/api.py", line 245, in launch_agent
    raise ChildFailedError(
```

[...]

PR #655, opened 2023-10-04
Only assert reshuffle if we are in train mode and we specify a data upsample refactor

humzaiqbal:
Currently the code will fail if you try to have multiple datasets with a resample factor because it will assert trying to resample erroneously even during test time. This PR makes sure we only have the assert statement during train time rather than test time. See this [issue](https://github.com/mlfoundations/open_clip/issues/632)


gabrielilharco:
Thanks @humzaiqbal!

PR #657, opened 2023-10-06
Add optional return for hidden states[SDXL compatibility]

nbardy:
Certain downstream tasks require operating on hidden states. 

Most notably SDXL is conditioned on these.

This is currently a draft, but I’d like to upstream this after it’s finished. Will be aiming to get the tests working in a few days. 

Wanted to open up discussion about the TransformerOutput class mostly.

Would be nice to have something standardized to return from CLIP’s with optional.


rwightman:
@nbardy so I understand the goals here, and it certainly makes it more HF transformers like w/ the output class, but there's a big problem in that it completely breaks compat for any existing users of OpenCLIP that are using either the tuple or dict outputs.

This could be achieved as an addon to the dict output (leaving tuple output for basic use), which would maintain compatiblity...

PR #662, opened 2023-10-07
NLLB-CLIP model

visheratin:
Hi.

I recently created NLLB-CLIP - a model that combines a text encoder from the [NLLB model](https://huggingface.co/facebook/nllb-200-distilled-1.3B) and an image encoder from the 
standard CLIP. This allows us to extend the model capabilities to 201 languages of the Flores-200. NLLB-CLIP sets state-of-the-art on the [Crossmodal-3600](https://google.github.io/crossmodal-3600/) dataset by performing very 
well on low-resource languages. You can find more details about the model in the [paper](https://arxiv.org/abs/2309.01859).

To integrate NLLB-CLIP to open_clip, I added the M2M100 config to HF configs, defined configs for the base and large models, trained two models using these configs, and added links to pre-trained models to the `pretrained.py` file.


gabrielilharco:
@visheratin do you have reference eval numbers for ImageNet or any of the 38 datasets we have at https://github.com/mlfoundations/open_clip/blob/main/docs/openclip_results.csv?

gabrielilharco:
I ran some evals and numbers look decent (> random) on many datasets. On ImageNet the base model gets 24.3% zero-shot acc, the large model gets 36.7%. On Flickr and COCO they look slightly different then on your table. Perhaps we used a different eval set. I also don't know if those are image or text retrieval.

rwightman:
k, figured out why mine failed, I was using the `hf-hub:visheratin/nllb-clip-base-oc` model, not local w/ pretrained arg ... the tokenizer in the HF hub is wrong, looks like CLIP tokenizer and not the nllb one... if loaded via the code w/o sourcing from the hub, i match 24% imagenet zs

PR #672, opened 2023-10-13
Add MetaCLIP models

gabrielilharco:
Add MetaCLIP models from Xu et al., 2023.

Paper: https://arxiv.org/abs/2309.16671
Code: https://github.com/facebookresearch/MetaCLIP

PS: this marks 100 models supported by OpenCLIP 🎉


ludwigschmidt:
💯

PR #690, opened 2023-10-22
move tok spacing from config to model init

gpucce:
This PR changes context length in coca cfg while shortening it in the model init so the tokenizer will behave as it should.

### DIFF
--- a/src/open_clip/coca_model.py
+++ b/src/open_clip/coca_model.py
@@ -23,21 +23,17 @@ try:
         RepetitionPenaltyLogitsProcessor,
         MinLengthLogitsProcessor,
         MaxLengthCriteria,
-        StoppingCriteriaList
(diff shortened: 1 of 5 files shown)


rwightman:
@gpucce can we not black the whole thing, I can't see what's changing, black is opinionated and we haven't been using it so that's a separate decision that needs to be done with a big PR that would update everything...

gpucce:
> @gpucce can we not black the whole thing, I can't see what's changing, black is opinionated and we haven't been using it so that's a separate decision that needs to be done with a big PR that would update everything...



> @gpucce can we not black the whole thing, I can't see what's changing, black is opinionated and we haven't been using it so that's a separate decision that needs to be done with a big PR that would update everything...

Sorry unintentional, will make a new PR

PR #692, opened 2023-10-23
Remove cls_embed arg from forward/encode_image fns

rwightman:
@gpucce wouldn't this be a cleaner approach to #691? the context_len (length of input text tokens) IS 76 so it doesn't make sense to feed it 77 and then always remove one token...


gpucce:
yeah it is clearly better

PR #693, opened 2023-10-23
Add torch flop counter to profiler

gabrielilharco:


rwightman:
@gabrielilharco with this, might be worth putting fvcore import in a try/catch so it'll run natively with torch if fvcore isn't installed (minus act count), and then assert if fvcore profile is selected and fvcore module not installed?

gabrielilharco:
Good call, will update

gabrielilharco:
Removing hierarchical flop counting for now until https://github.com/pytorch/pytorch/issues/111809 is resolved. With https://github.com/mlfoundations/open_clip/pull/699 it now works for all supported models

rwightman:
@gabrielilharco looks good, do you want to merge #693 #699 #701 to main and release?

gabrielilharco:
Sounds good!

PR #695, opened 2023-10-23
Fix `model.get_model_preprocess_cfg`

bryant1410:
For this PR, I'm not 100% sure it's a bug but it feels like it is. This function first checks there's an "image_mean" attribute but then gets a different one ("mean"). Similarly for "std". So it fixes this.


rwightman:
@bryant1410 woops, yeah, missed second step of a change

PR #696, opened 2023-10-23
Move CLIPA weights to UCSC-VLAA org

rwightman:


rwightman:
@zw615 the weights are moved to your org. When you edit the README, should avoid changing the metadata in the top section (part between the two `---`) such as library tags etc, as it can break hub functionality.

zw615:
@rwightman Copy that. Thanks Ross!

PR #679, opened 2023-10-17
Release 2.23.0

gabrielilharco:
Closes #678

PR #741, opened 2023-11-18
NLLB-CLIP with SigLIP + small tokenizer fix

visheratin:
Hi! I trained NLLB-CLIP models with SigLIP (ViT and loss). They perform much better than the previous version across all benchmarks.

I'm also working on integrating the multilingual benchmarks from the [paper](https://arxiv.org/abs/2309.01859) into the [CLIP benchmark](https://github.com/LAION-AI/CLIP_benchmark). To make it work with the NLLB tokenizer, I had to change the tokenizer method to `batch_encode_plus` because the default `__call__` doesn't take language-specific prefix tokens into account.


rom1504:
Could you share results? They would also fit well in the readme or its
referenced files

rom1504:
I think it would be interesting to see how it compares to vit-h/14
multilingual in this repo

yes it could make sense to have a different csv for multilingual models
indeed

rom1504:
xlm-roberta-large-ViT-H-14

rom1504:
thanks, interesting indeed!

The PR LGTM

I think adding a mention on this new model in this section https://github.com/mlfoundations/open_clip/blob/main/docs/PRETRAINED.md#nllb would be good, so people have a change to discover the model (by looking at openclip doc)

rom1504:
this is how I trained this xlm-roberta-large-ViT-H-14 model https://github.com/mlfoundations/open_clip/blob/main/docs/PRETRAINED.md#vit-h14-xlm-roberta-large

looking at your paper https://arxiv.org/pdf/2309.01859.pdf the freezing method seems a bit similar; I froze the image encoder but not the text encoder
However I trained on very different data: noisy multilingual captions from Laion5B (VS the much cleaner but smaller dataset you used)

I see you evaluate only on retrieval. I had evaluated as well on imagenet with translated class names, and found the model to perform better than previous ones but in absolute numbers really poorly. (for example 56% in italian while the model gets 78% for the same in english). I am not sure what is the cause but that may be of interest to you

(just FYI)

gabrielilharco:
@visheratin looks from my end too. My scripts for running evals on the 38 datasets still need some cleaning up, I plan to do that in the future and push them here so everyone can run it easily. Meanwhile, I'm running evals for the 2 new models, and will update here with the results once it's done

gabrielilharco:
@visheratin I added eval results and profiling numbers for the new models

gabrielilharco:
Thanks @visheratin! Can you make the numbers in the numbers in the new csv have less significant digits? E.g.  0.8569999933242798 becomes 0.8570. I think it's a bit easier to read this way.

It would be nice to add the other nllb models to that table as well. Ideally all models actually, would this be too expensive for you to run? If so, I can try running on my end if you share the scripts.

gabrielilharco:
Sounds good to me. Thanks @visheratin!

BIGBALLON:
Hi, @visheratin , thanks for your great work, Is there any plan to add NLLB-CLIP models(with SigLIP) to [timm](https://github.com/huggingface/pytorch-image-models)

BIGBALLON:
@visheratin yea, thanks, looking forward to the next verison of open_clip.

rom1504:
You can make a PR like this if you want a release
https://github.com/mlfoundations/open_clip/pull/679
That creates a pipy release on merge

PR #756, opened 2023-12-07
Add `get_logits` method and NLLB tokenizer

visheratin:
Hi!

I want to make OpenCLIP more usable for downstream applications, like zero-shot classification. Right now, to get the logits, the user has to call `encode_image` and `encode_text`, then matmul them, multiply the result by `logit_scale`, and optionally add `logit_bias`. I think it makes sense to have one method to get logits, as in OpenAI and HuggingFace. So I added the `get_logits` method to both `CLIP` and `CustomTextCLIP` classes.

I also added the `NLLBTokenizer` class that has an additional `langs` parameter in the `__call__` method. This is needed because the tokenizer for NLLB models adds a language token to the beginning of the sequence. The token that is added is controlled via the `set_src_lang_special_tokens` method. If the language is not set via this method, the tokenizer will add an English token to all sequences.

The PR also contains some formatting changes performed by Ruff. For me personally, the formatted code looks nicer, but if you don't like it, I can roll it back.

@gabrielilharco @rwightman

### DIFF
--- a/src/open_clip/factory.py
+++ b/src/open_clip/factory.py
@@ -9,58 +9,82 @@ from typing import Any, Dict, Optional, Tuple, Union
 
 import torch
 
-from .constants import OPENAI_DATASET_MEAN, OPENAI_DATASET_STD
-from .model import CLIP, CustomTextCLIP, convert_weights_to_lp, convert_to_custom_text_state_dict,\
-    resize_pos_embed, get_cast_dtype, resize_text_pos_embed, set_model_preprocess_cfg
 from .coca_model import CoCa
-from .loss import ClipLoss, DistillClipLoss, CoCaLoss, SigLipLoss
+from .loss import ClipLoss, CoCaLoss, DistillClipLoss, SigLipLoss
+from .model import (
+    CLIP,
+    CustomTextCLIP,
+    convert_to_custom_text_state_dict,
+    convert_weights_to_lp,
+    get_cast_dtype,
+    resize_pos_embed,
+    resize_text_pos_embed,
+    set_model_preprocess_cfg,
+)
 from .openai import load_openai_model
-from .pretrained import is_pretrained_cfg, get_pretrained_cfg, download_pretrained,\
-    list_pretrained_tags_by_model, download_pretrained_from_hf
-from .transform import image_transform_v2, AugmentationCfg, PreprocessCfg, merge_preprocess_dict, merge_preprocess_kwargs
-from .tokenizer import HFTokenizer, SimpleTokenizer, DEFAULT_CONTEXT_LENGTH
+from .pretrained import (
+    download_pretrained,
+    download_pretrained_from_hf,
+    get_pretrained_cfg,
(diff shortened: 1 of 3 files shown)


rwightman:
@visheratin can we do this without blacking it and changing more lines than being added?

rwightman:
also, should probably be rebased against latest

src/open_clip/model.py
```diff
@@ -285,6 +285,15 @@ def encode_text(self, text, normalize: bool = False):
...
 
+    def get_logits(self, image, text):
+        image_features = self.encode_image(image, normalize=False)
+        text_features = self.encode_text(text, normalize=False)
+        image_logits = self.logit_scale * image_features @ text_features.T
+        if self.logit_bias is not None:
+            image_logits += self.logit_bias
+        text_logits = image_logits.T
+        return image_logits, text_logits
+
```
rwightman, line 296: This should be 
```python
    def get_logits(self, image, text):
        image_logits = self.logit_scale.exp() * image_features @ text_features.T
        if self.logit_bias is not None:
            image_logits += self.logit_bias
        text_logits = image_logits.T
        return image_logits, text_logits
```

src/open_clip/factory.py
```diff
@@ -111,11 +110,18 @@ def get_tokenizer(
         context_length = text_config.get('context_length', DEFAULT_CONTEXT_LENGTH)
 
     if 'hf_tokenizer_name' in text_config:
-        tokenizer = HFTokenizer(
+        if model_name.startswith("nllb"):
+            tokenizer = NLLBTokenizer(
```
rwightman, line 114: really not a fan of having a model name based hack

rwightman:
So, for the tokenizer, I'm not convinced it warrants a new tokenizer and associated maintenance. Isn't it pretty standard for multi-lingual to manually insert the language token per text? Are there any popular impl which do it this way?

rwightman:
On the get_logits, it's useful to have, but will point out this won't work with torchcompile or FSDP which only wrap forward() methods.

rwightman:
@visheratin rather than make a whole new tokenizer for this with a different tokenize interface, couldn't we pass through the set_src_lang methods and/or src lang init kwargs? assert/report error if it's called on a underlying HF tokenizer that doesn't have it?

rwightman:
> Passing through the `set_src_lang` method is an option. But it will still be non-obvious for the user that they may need to use this method if they want to use languages other than English. Also, if we go this way, we make the users implement the logic that is now in the `NLLBTokenizer` themselves if they have multilingual input.
> 
> Maybe multilingual input texts is too edge case. If you think so, I can remove `NLLBTokenizer` from the PR. If the user wants to change the source language, they can always call `tokenizer.tokenizer.set_src_lang`, as it is done in the CLIP benchmark.

Isn't this how the HF tokenizers work though? I think they have some sort of specific tokenization method with src/target lang as args, but usually it's either set on construction of the tokenizer or via the set method no?

PR #767, opened 2023-12-20
Update params.py: "BatchNorm" -> "LayerNorm" in description string

xwen99:
"BatchNorm" -> "LayerNorm" in the description string for "--lock-text-freeze-layer-norm"


gabrielilharco:
Thanks @xwen99

PR #778, opened 2023-12-30
Drop hard dependency on `timm`

akx:
All uses of `timm` are already (and have been) late or guarded by import checks, so it shouldn't be a hard dependency.


rwightman:
too many models depend on timm now so not a good idea, timm is pretty minimal as far as deps go

PR #784, opened 2024-01-07
Release 2.24.0

rom1504:


rom1504:
any reason not to release @gabrielilharco @rwightman ?

gabrielilharco:
Not from my side. One thing, there's a new bigG metaclip that would be good to add here eventually, but I haven't had the time to run evals on it yet. Just an fyi, don't think it's a blocker

rwightman:
@rom1504 yeah, should be good, still have broken coca training was conflicted on how to proceed, with another sorta meh approach or try and improve more fundamentally but last release was same in that regard so go for it...

rwightman:
I never remember whether it's rebase or squash that messes up the release process, so you can kick it off @rom1504

PR #820, opened 2024-02-18
safetensors support when loading from hf_hub

deniz-birlikci:
**Changes**

- requirements file includes safetensors import
- load_checkpoint function now loads safetensors if there's .safetensors suffix
- download_pretrained_from_hf function checks for .bin file first, and if that isn't found, checks for .safetensors file

**Why is this important?**
With safetensors being much faster in loading up the weights, this is a helpful change for the repository. These modifications only come into effect if the HF repository does not have the .bin file, therefore, there would be no changes to the current open_clip users. My changes only come into effect if the .bin file is missing, and instead of raising an Exception, it checks the repo again for .safetensors file before doing so.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,7 @@ regex
 ftfy
 tqdm
 huggingface_hub
+safetensors
 sentencepiece
 protobuf
 timm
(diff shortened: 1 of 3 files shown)


rwightman:
merged in parent pr, but forced to squash

PR #860, opened 2024-04-12
Fix stopping_criteria result check in coca_model

### LINKED ISSUE #847: RuntimeError: Boolean value of Tensor with more than one value is ambiguous
The coca model demo not work and the wrong code is here:
```py
# open_clip/coca_model.py 
...
if beam_scorer.is_done or stopping_criteria(input_ids, None):
  break
...
```

the beam_scorer.is_done is like: ensor(False, device='cuda:0')
the stopping_criteria(input_ids, None) is like:  tensor([False, False, False, False, False, False], device='cuda:0')

the beam_scorer.is_done or stopping_criteria(input_ids, None) is like tensor([False, False, False, False, False, False], device='cuda:0')
but when you run `if tensor([False, False, False, False, False, False], device='cuda:0') ` it will case RuntimeError: Boolean value of Tensor with more than one value is ambiguous

MengqingCao:
**fix #847**

The stopping criteria is updated in the latest `transformers`(V4.39.3 now). The return result is modified to a tensor (`torch.full((input_ids.shape[0],), is_done, device=input_ids.device, dtype=torch.bool)` ) instead of a bool value, which causes the bug in #847

**the related code in `transformers`**
https://github.com/huggingface/transformers/blob/0bd58f1ce0573c0e3269de4215a17d318add49b9/src/transformers/generation/stopping_criteria.py#L76

### DIFF
--- a/src/open_clip/coca_model.py
+++ b/src/open_clip/coca_model.py
@@ -23,6 +23,8 @@ try:
         RepetitionPenaltyLogitsProcessor,
         MinLengthLogitsProcessor,
         MaxLengthCriteria,
+        StopStringCriteria,
+        EosTokenCriteria,
         StoppingCriteriaList
     )
 
@@ -298,7 +300,12 @@ class CoCa(nn.Module):
 
                 cur_len += 1
 
-                if stopping_criteria(out, None).any():
+                is_done = False
+                if EosTokenCriteria in stopping_criteria or StopStringCriteria in stopping_criteria:
+                    is_done = stopping_criteria(out, None).all()
+                else:
+                    is_done = stopping_criteria(out, None).any()
+                if is_done:
                     break
 
             if num_dims == 1:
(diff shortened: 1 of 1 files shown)


MengqingCao:
cc @rwightman @rom1504 @gabrielilharco @bryant1410 @mitchellnw

rwightman:
hmm, yeah this needs fixing. One q though as I'm not intimately familiar with the gen code. Does the implementation here support multiple sentences in a batch? If yes, using `any()` will stop too early as the bool tensor is done state for each row (sentence) no?  If it's only supporting one sentence at a time this is fine for now..
@gpucce ?

MengqingCao:
> @MengqingCao hmm, yeah this needs fixing. One q though as I'm not intimately familiar with the gen code. Does the implementation here support multiple sentences in a batch? If yes, using `any()` will stop too early as the bool tensor is done state for each row (sentence) no? If it's only supporting one sentence at a time this is fine for now.. @gpucce ?

Your concerns are right in the cases when users using `StopStringCriteria` and `EosTokenCriteria`, which I ignored before. I only noticed the default `StoppingCriteria` method `MaxLengthCriteria` before, which returns a boolTensor filled with one single bool value `is_done`. Thus, I think use `any()` brings bigger operating efficiency than `all()`.

[The related code ](https://github.com/huggingface/transformers/blob/47735f5f0f2752500d115d2f6bd57816032599b6/src/transformers/generation/stopping_criteria.py#L83) in Transformers:
image

To adapt to the situation of `StopStringCriteria` and `EosTokenCriteria` at the same time, I think we have two choices:
1. change to use `all()` here
2. checking if there is `StopStringCriteria` and `EosTokenCriteria` in `stopping_criteria`, if no, use `any()`, otherwise, use `all()`. This may run faster but bring more changes than 1

MengqingCao:
@rwightman @gpucce , I have implemented option 2 and updated the code, give me some suggestions plz, thanks!

PR #883, opened 2024-05-21
load OpenCLIP model and tokenizer from local path

YushunXiang:
Load OpenCLIP model and tokenizer from local path

Usage:

``` python
model, processor = open_clip.create_model_from_pretrained(
    model_name=f'hf-hub:{model_id}',
    precision='fp16',
    cache_dir=cache_dir,
)
print(f"Loading tokenizer from {model_args.model_path}")
tokenizer = open_clip.get_tokenizer(
    model_name=f'hf-hub:{model_id}',
    cache_dir=cache_dir,
)
```

The `cache_dir` is your local path.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -151,3 +151,4 @@ core.*
 
 # Allow
 !src/evaluation/misc/results_dbs/*
+src_copy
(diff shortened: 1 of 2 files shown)


rwightman:
so this will interfere with normal use of the cache by overriding part of the functionality. I've made some improvements to the cache operation. Explicitly loading from local folders is a bit of a different idea and I'd like to support that but should be done through pretrained argument (folder) and not by overriding cache behaviour

PR #967, opened 2024-10-21
Add safeglobals to allow metaclip models to load with weights_only=True

rwightman:


src/open_clip/factory.py
```diff
@@ -26,6 +26,20 @@
 _MODEL_CONFIGS = {}  # directory (model_name: config) of model architecture configs
 
 
+try:
+    import _codecs
+    import numpy as np
+    # add safe globals, known to be needed for metaclip weights
+    torch.serialization.add_safe_globals([
+        _codecs.encode,  # now in pytorch main but some pytorch versions w/ weights_only flag don't have it
```
bryant1410, line 34: What if you specify that's not necessary from PyTorch 2.5.0 (I checked this)?

bryant1410: For accountability, this is how I checked it's in that release (tag): I opened https://github.com/pytorch/pytorch/commit/1e9bedf6880f813b165e357cd62c8a7238789aad, and saw that it contains the tag v2.5.0 (meaning that this commit is included in the tag, thus the release).

Issue #771, opened 2023-12-23
Error-prone behavior of supporting the "openai" checkpoint with non-QuickGELU models


bryant1410:
Hey, I believe that supporting to run "openai" pre-trained checkpoint with non-QuickGELU models (e.g., `RN50` and `ViT-B-32`) leads to bugs. The pattern is the following:

1. Fine-tune an OpenAI-pre-trained CLIP model (e.g., with `--model ViT-B-32 --pretrained openai`) inadvertently using a non-QuickGELU model, which is fine because [it's hardcoded to use QuickGELU](https://github.com/mlfoundations/open_clip/blob/73ad04ae7fb93ede1c02dc9040a828634cb1edf1/src/open_clip/factory.py#L201) anyway.
2. Use the same command to run the evaluation but change `openai` for the fine-tuned model path.

What will happen is that the native GELU will be used instead of QuickGELU (the latter was used to train the model), and wrong results will be obtained.

This happened to me, as well as others (though there's a pending confirmation from them):

* mertyg/vision-language-models-are-bows#35 (ARO/NegCLIP paper)
* RAIVNLab/sugar-crepe#7 (SugarCrepe paper)

Would it be possible to fix/avoid this error-prone pattern? I see some ways:

* Not allowing the running of non-QuickGELU models with the "openai" pre-trained checkpoint. Maybe it can be detected with a special error message (pointing to this issue), inviting the user to use the correct model variant.
* Give a warning to the user when "openai" is used without the QuickGELU model (or without the `--force-quick-gelu` flag).

PR #970, opened 2024-10-23
All default pretrained weights pushed to HF hub

rwightman:
* all weights not already on the hub uploaded to timm org for simplicity & safety (metaclip models were incompatible with weights_only=True).
* OpenAI models no longer use special path that loads from torchscript archive, use same path as other models
* Handling of QuickGELU consistent btw openai and non, made a bit more safe, warn on mismatch
* safetensors is default weight load if available

### DIFF
--- a/src/open_clip/factory.py
+++ b/src/open_clip/factory.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 import re
+import warnings
 from copy import deepcopy
 from dataclasses import asdict
 from pathlib import Path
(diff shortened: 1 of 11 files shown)


bryant1410:
IIUC, this fixes #771.

bryant1410 (approved):
Look at the core changes and it LGTM, though haven't looked at all the details.

This is a great and elegant solution for both the QuickGELU error-prone issue, and the issue with MetaCLIP!!

PR #971, opened 2024-10-26
Experimenting with alternative siglip loss impl for better dist scaling

rwightman:
@JeniaJitsev @mehdidc


JeniaJitsev:
Very nice, giving it a try

rwightman:
So gather impl here looks like it's a better balance but still scaling issues wrt to throughput. The additions here allow flexiblility to add other impl that can further improve things so will merge what's here for now.

PR #1008, opened 2024-12-13
Add CLIPS to open_clip

Yanqing0327:
Add the CLIPS model, which is trained on public Recap-DataComp-1B.
Performance:
image

### DIFF
--- a/src/open_clip/pretrained.py
+++ b/src/open_clip/pretrained.py
@@ -318,9 +318,9 @@ _VITH14 = dict(
         hf_hub="timm/vit_huge_patch14_clip_224.metaclip_2pt5b/",
         quick_gelu=True,
     ),
-    metaclip_alltogether=_pcfg(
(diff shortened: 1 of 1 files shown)


rwightman:
@Yanqing0327 there's already a config based mechanism to override arguments to the normalization layer, 

In the image and vision configs you can add
```
        "norm_kwargs": {
            "eps":  1e-6
        }
```

I'm also not quite clear on what the extra logic in the different tokenizer path involves, is there no way to use the existing HFTokenizer wrapper with some minimal pre/post processing that can be enabled via config?

Yanqing0327:
Thank you so much for your help! I’ve already added the eps to the config as you suggested. However, for the tokenizer part, I haven’t found a cleaner implementation yet.

rwightman:
@Yanqing0327 after that last change, this PR is out of sync with main, it's reverting a bunch of recent changes.

Issue #1049, opened 2025-03-14
Deprecated usage in README example


Ja-Tink:
In the sample code in the README to create an embedding of the sample dog image a deprecated PyTorch function (`torch.cuda.amp.autocast(args...)`) is used. See screenshot.

PyTorch Version: 2.6.0
OpenClip Version: 2.31.0

<img width="1188" alt="Image" src="https://github.com/user-attachments/assets/e5730d11-0ccc-46ec-a418-4bf83668308c" />

rwightman:
merged PR, thanks

PR #1050, opened 2025-03-14
Updated PyTorch code in README 

Ja-Tink:
Updated README sample code to address #1049


rwightman:
@Ja-Tink ... just `torch.autocast` is the more common use now isn't it? without the extra .amp namespace, seems to be the preferred one in docs

Ja-Tink:
Yes, you're right. I should've checked the docs instead of just following the warning. Thanks for pointing that out! I've updated the PR to reflect this.

Issue #1068, opened 2025-04-12
SigLIP 2 offline fails silently - loads wrong img preprocessor


guillaumeguy:
I have seen folks try to load the SigLIP 2 offline. #1056 and #1053 . You have to be quite careful as it seems to use a different image preprocessor which is a big footgun. The resulting embeddings are quite different from what I have seen.

@rwightman proposes a different approach (I used `huggingface_hub.snapshot_download` in my approach)
https://github.com/mlfoundations/open_clip/issues/1056#issuecomment-2741001726

![Image](https://github.com/user-attachments/assets/4e340007-e224-4784-a756-1046cc695f01)

reproducible code:
```python
# From hub
from open_clip import create_model_from_pretrained, get_tokenizer # works on open-clip-torch >= 2.31.0, timm >= 1.0.15
create_model_from_pretrained('hf-hub:timm/ViT-SO400M-16-SigLIP2-384')[1]

# Local (download first)
name = 'ViT-SO400M-16-SigLIP2-384'
import open_clip
open_clip.create_model_from_pretrained(model_name=name, pretrained='/tmp/ViT-SO400M-16-SigLIP2-384/open_clip_model.safetensors')

rwightman:
@guillaumeguy maybe not obvious, but when you load just the pth file like that you aren't using the same model config, you're just loading the weights specified into the model architecture specified and it uses default image processing.

If you want to cache for offline should probably use cache_dir ... don't have a load from folder w/ config file yet.

guillaumeguy:
> model architecture

Ah. interesting. I was convinced that the tokenizer + image preprocessor were a function of the model name, not the weights passed (as a user, I am surprised the weights could change the first 2).

Thanks for your quick reply! 

I know you already talked about it in the other thread but I would reinforce the nicety of having an "offline option" for model initialization + weight loading .

rwightman:
@guillaumeguy because of the original setup, it's a bit confusing ... the builtin model configs (https://github.com/mlfoundations/open_clip/tree/main/src/open_clip/model_configs) cover the architecture, and does incl the tokenizer as it tighly coupled with the text model archiecture used), but the image preprocessing was associated with the weights you can see the matching of pretrained weights here https://github.com/mlfoundations/open_clip/blob/main/src/open_clip/pretrained.py 

[...]

PR #1069, opened 2025-04-16
Initial work on adding local-dir: schema for model & tokenizer loading from local folder

rwightman:
Also
* add support to load specific image / text tower weights
* refine consistency for image (timm), text (hf) pretrained weight loading on init Currently lots of logs while testing

### DIFF
--- a/src/open_clip/factory.py
+++ b/src/open_clip/factory.py
@@ -404,7 +404,7 @@ def create_model(
                     # QuickGELU compatibility check will happen in after force overrides
                 except Exception as e:
                     logging.error(f"Failed to download weights for tag '{pretrained}': {e}")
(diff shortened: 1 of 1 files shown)


rwightman:
Confirmation of functionality from users would be helpful... 

Can use `local-dir:path/to/folder` in `create_model_*` and `get_tokenizer` to load models from a folder that contain a HF hub compatible open_clip_config.json and associated .safetensor or .bin file...

guillaumeguy:
FYI: Looks good to me anecdotally: 

https://github.com/mlfoundations/open_clip/issues/1068#issuecomment-2830733946

elonmusxk-dev:
🚀

PR #1086, opened 2025-06-09
Wire up custom attention block via config

rwightman:
add qk_norm, and enable qk_norm, scaled_cosine_attn, scale_heads, attn and inner attn norm scaling via config

### DIFF
--- a/src/open_clip/transformer.py
+++ b/src/open_clip/transformer.py
@@ -264,6 +264,9 @@ class ResidualAttentionBlock(nn.Module):
         ]))
         self.ls_2 = LayerScale(d_model, ls_init_value) if ls_init_value is not None else nn.Identity()
 
+    def get_reference_weight(self):
+        return self.mlp.c_fc.weight
+
     def attention(
             self,
             q_x: torch.Tensor,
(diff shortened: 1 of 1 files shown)


rwightman:
@JeniaJitsev sounds good, there's ideas from diff papers there and they wouldn't necessarily all make sense in combo ...

qk_norm and scaled_cosine_attn are explicitly disabled together, does not make sense

but also using qk_norm in combo with inner norm probably not worthwhile

using scale_head + inner norm together not worthwhile

scale head + scale_attn ('outer' norm) was part of normformer so works well together

scale fc (norm in mlp) with any of the attn focused options could be worthwhile

rwightman:
I've found layer scale to benefit quite a few vit training regimes... that's been supported for awhile but not sure if any from-scratch runs use it.... ls init values usually set btw 1e-5 and 0.1

JeniaJitsev:
@rwightman  So, it seems it makes sense to test 2 setups:
1. qk norm active (otherwise everything else standard training)
2. scale head + scale_attn (as already employed in normformer)

Hope 1) and 2) I can get by having a corresponding json conf for each setup without further source code changes in this PR. Will check then with you before starting training whether confs make sense.

PR #1105, opened 2025-08-06
Support loading a model in a non-init-weights context

bryant1410:
I guess this is a borderline bug fix/feature addition. There's the common use of using context manager that sets the model in the meta device by default when transitively loading a model (such as `accelerator.init_empty_weights` or [what's used in transformers by default for `from_pretrained` since a couple of months ago](https://github.com/huggingface/transformers/blob/ac0b4684657cbff1e8eecb0d966d10b71843dca0/src/transformers/modeling_utils.py#L4988). The problem is that `open_clip.factory.create_model` moves the model to a device, and thus it's incompatible for this. I added a check for this case and a log message for it.

This is a code block to reproduce this error (it works fine after the fix):

```python
import open_clip
import transformers

class Config(transformers.PretrainedConfig):
    pass

class Model(transformers.PreTrainedModel):
    config_class = Config

    def __init__(self, config):
        super().__init__(config)
        self.another_model = open_clip.create_model_and_transforms("ViT-B-32-quickgelu", pretrained="openai")[0]


a = Model(Config())
a.save_pretrained("/tmp/abc")

b = Model.from_pretrained("/tmp/abc")
```


rwightman:
To provide more detail on my thoughs, I have some ideas to impl a two-phase init approach for OpenCLIP / timm, where the factories will allow the model to be created on meta device (under appropriate context) w/ no weight-init, and then can either load weights or trigger final init via a consistent `_init_weights` / `init_weights` call if no pretrained loaded.

Now, in your use case it looks like you're using an OpenCLIP model as a submodule in a larger one that has a transformers base. So, already had this hiccup running through scenarios. If a timm or OpenCLIP model factor is called from within a parent context, we need to detect and adapt the two-phase approach assuming that the parent will be taking the final action re weight loading or triggering init. Does that make sense or did I get the intention in your use case wrong?

rwightman:
@bryant1410 yeah, I got that you're loading the weights outside of OpenCLIP and are already in a meta device context with the hacky patchy approach. 

I was thinking about both making OpenCLIP and timm models meta device compatible (timm especially is not right now) AND adding a two phase init approach mentioned in the hopes of avoiding the monkey patching altogether as I *REALLY* don't like that approach.

I'd like any interm modifications like this one to be compatible with the end approach, and not require possibly breaking changes...  your proposed changes are fairly minimal but need to run through if there's any unforseen consequence

rwightman:
I've started working on better init schemes for timm, and eventually here... in the mean time, taking another peak at this, I think this would cover this use case without adding too much logic that would be invalidated by future changes


```
    model_is_in_meta_device = next(model.parameters()).device.type == "meta"
    if not model_is_in_meta_device:
        _set_model_device_and_precision(model, device, precision, is_timm_model)
        model_is_in_meta_device = device.type == 'meta'
```

And then have have the branch for a non-warning log that model is uninitialized / on the meta device.

I'd prefer to leave the checkpoint loading logic alone. `load_weights=False` should disable loading the weights and there will be future code where meta device use will involve materialization and loading in the factory.

FWI the code does currently work fine if you do 

```
with torch.device('meta'):
   self.another_model = open_clip.create_model_and_transforms("ViT-B-32-quickgelu", pretrained="openai", device="meta", load_weights=False)[0]

actually works without the load_weights too

Issue #942, opened 2024-09-26
SigLip memory consumption increases as we scale number of GPUs


khalidsaifullaah:
From the SigLip paper my understanding is that it doesn't require any `all_gather` and it's always performing local `b x b` computation iteratively, where b is `micro_batch_size` (see this section from the paper). 
![image](https://github.com/user-attachments/assets/e5e2dbc8-c4a5-4a66-991e-7ca4266a6bd5)

So if I can fit let's say `micro_batch_size` 10 (in 8 GPUs), and then I increase the number of GPUs to 16, 32, 64, 128, ... my memory consumption should (more or less) remain the same (just like doing normal DDP). Or simply put, we should be able to scale `world_batch_size` or the number of nodes by keeping the `micro_batch_size` constant (in theory) right?

But what I've observed is that the memory consumption spikes as i increase `world_batch_size` (num of nodes) and I need to lower my `micro_batch_size` (even to as low as `2` for 128 devices).

1. I'm wondering if my understanding of siglip is correct that keeping the `micro_batch_size` constant it allows you to scale `world_batch_size`? It could also be the case that they do some sort of TPU trick?
2. I have only skimmed through the [siglip implementation here](https://github.com/mlfoundations/open_clip/blob/fc5a37b72d705f760ebbc7915b84729816ed471f/src/open_clip/loss.py#L307) and I think it could also be possible that while swapping neighbors it doesn't not free up the memory, and that's why the consumption accumulates...?

I could be totally wrong on both of these, so I'd be glad to know if anyone tried scaling `world_batch_size` and have had similar results, so i could validate my hypothesis

rwightman:
@khalidsaifullaah yeah, it's not working quite as efficiently as it should. I feel my current isend/irecv impl, while in theory should be reasonable, it appears it may not a well optimized approach. 

Looking at the big_vision codebase where SigLIP authors have the original ver of the models, there's no chunked sigmoid impl in the current code, but there's an impl in a deprecated file, interestingly there's a comment that states the 'ppermute' version (which should be equivalent to send/recv neighbour exchange, and should itself be more optima) used more memory than doing 'hot potatoe' with all_reduce. Hmmm https://github.com/google-research/big_vision/blob/46b2456f54b9d4f829d1925b78943372b376153d/big_vision/trainers/proj/image_text/_deprecated_contrastive.py#L168-L200

[...]

Issue #765, opened 2023-12-15
SigLIP memory issue with a large batch size


airogachev:
Original paper clamed to use big batches.
Using current implementation I face the problem that if I increase the batch size even to 1024, it fails on the **second** iteration. 
I use 4 cards with 44.5 GB of video memory. So, it seems that memory may be filled with something that was not cleaned after the first batch? Any ideas? 

```   
File "/opt/code-server/lib64/python3.8/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
    return forward_call(*input, **kwargs)
  File "/opt/code-server/lib64/python3.8/site-packages/torch/nn/parallel/distributed.py", line 1040, in forward
    output = self._run_ddp_forward(*inputs, **kwargs)
  File "/opt/code-server/lib64/python3.8/site-packages/torch/nn/parallel/distributed.py", line 1000, in _run_ddp_forward
    return module_to_run(*inputs[0], **kwargs[0])
  File "/opt/code-server/lib64/python3.8/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
    return forward_call(*input, **kwargs)
  File "/home/jovyan/itm-v2/open_clip/src/open_clip/model.py", line 254, in forward
    text_features = self.encode_text(text, normalize=True) if text is not None else None
  File "/home/jovyan/itm-v2/open_clip/src/open_clip/model.py", line 241, in encode_text
    x = self.transformer(x, attn_mask=self.attn_mask)
  File "/opt/code-server/lib64/python3.8/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
    return forward_call(*input, **kwargs)
  File "/home/jovyan/itm-v2/open_clip/src/open_clip/transformer.py", line 321, in forward
    x = r(x, attn_mask=attn_mask)
  File "/opt/code-server/lib64/python3.8/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
    return forward_call(*input, **kwargs)
  File "/home/jovyan/itm-v2/open_clip/src/open_clip/transformer.py", line 243, in forward
    x = x + self.ls_2(self.mlp(self.ln_2(x)))
  File "/opt/code-server/lib64/python3.8/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
    return forward_call(*input, **kwargs)
  File "/opt/code-server/lib64/python3.8/site-packages/torch/nn/modules/container.py", line 204, in forward
    input = module(input)
[...]

PR #1145, opened 2026-04-05
Add chunked logit computation to SigLipLoss for memory efficiency

avocardio:
## Summary

Adds a `chunk_size` parameter to `SigLipLoss` that chunks the B×N logit matrix computation, reducing peak memory from O(B×N) to O(chunk_size×N). This addresses the memory scaling issues reported in #942 and #765.

**Usage:**
```python
loss_fn = SigLipLoss(chunk_size=512)  # chunks logits into 512-row blocks
```

Default `chunk_size=0` preserves the original behavior exactly — no changes to existing code paths.

All four distributed strategies (bidir, shift, reduce, gather) benefit automatically since they all call `_loss()` internally.

## Memory savings

At B=4096, D=768:
- Original: 64MB for logit matrix
- chunk_size=512: 8MB (8× less)

At B=32768 (SigLIP's optimal batch size):
- Original: 4GB for logit matrix
- chunk_size=512: 64MB (64× less)

## Changes

- Added `chunk_size` parameter to `SigLipLoss.__init__`
- Added `_chunked_loss()` method that chunks over image feature rows
- `_loss()` dispatches to `_chunked_loss()` when `chunk_size > 0`
- No new dependencies, pure PyTorch
- Zero changes to the default code path

## Related issues

- #942 — SigLip memory consumption increases with GPU count
- #765 — OOM with SigLIP at batch_size=1024

### DIFF
--- a/src/open_clip/loss.py
+++ b/src/open_clip/loss.py
@@ -395,19 +395,18 @@ class SigLipLoss(nn.Module):
 
         for i in range(0, B, chunk_size):
             end_i = min(i + chunk_size, B)
+            c = end_i - i
             img_chunk = image_features[i:end_i]
-            logits = self.get_logits(img_chunk, text_features, logit_scale, logit_bias)
-            labels = self.get_ground_truth(
-                image_features.device,
-                image_features.dtype,
-                logits.shape[0],
-                negative_only=True,  # start with all negative
-            )
+            logits = self.get_logits(img_chunk, text_features, logit_scale, logit_bias)  # (c, N)
+
+            # Build (c, N) rectangular label tensor: -1 everywhere, +1 at the global diagonal.
+            # We cannot use get_ground_truth here because it returns a square (c, c) tensor.
+            labels = -torch.ones((c, N), device=logits.device, dtype=logits.dtype)
             if not negative_only:
-                # Set diagonal: positive pair at (local_row, global_col) where global_col == i + local_row
-                for k in range(end_i - i):
-                    if i + k < N:
-                        labels[k, i + k] = 1.0
+                cols = torch.arange(i, min(end_i, N), device=logits.device)
+                rows = torch.arange(cols.shape[0], device=logits.device)
+                labels[rows, cols] = 1.0
+
             total_loss = total_loss + (-F.logsigmoid(labels * logits).sum())
 
         return total_loss / B
--- a/src/open_clip/loss.py
+++ b/src/open_clip/loss.py
@@ -387,27 +387,35 @@ class SigLipLoss(nn.Module):
 
         Peak memory: O(chunk_size * N) instead of O(B * N).
         Useful when per-device batch is large (e.g. B > 4096).
+
+        Uses the identities -logsigmoid(-x) = softplus(x) and
+        softplus(-x) - softplus(x) = -x to avoid materializing a labels
+        tensor: the all-negative loss is softplus(logits), and each diagonal
+        positive needs only a -logits[k, i+k] correction.
         """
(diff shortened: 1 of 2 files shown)


avocardio:
## Memory savings across GPU architectures

Tested at D=768, fwd+bwd, f32. Peak memory savings from chunked logit computation:

| Batch | Reference | Chunked (cs=512) | Reduction |
|-------|-----------|------------------|-----------|
| 16,384 | 4.2 GB | 288 MB | **14.7×** |
| 32,768 | 16.7 GB | 576 MB | **28.9×** |
| 65,536 | 66.1 GB | 1.15 GB | **57×** |

**Speed:** this is a pure-PyTorch loop, so the per-chunk launch overhead means wall time is comparable to or slightly slower than the unchunked path; expect no speedup. The win is enabling batch sizes that previously OOM'd. For users who also want a speedup, we maintain a Triton-kernel variant at https://github.com/avocardio/siglip-kernel that fuses the matmul and sigmoid BCE in a single pass.

rwightman:
@avocardio so I took a look at this, it doesn't appear to work?

avocardio:
@rwightman Sorry, fixed in 357bd97 — added a parity test.

rwightman:
@avocardio k, that's working now... but, FWIW when I was looking at this I ran it by code and it suggested a further level of optimization that reduces allocations and should be equivalent down to floating point rounding


```
  def _chunked_loss(self, image_features, text_features, logit_scale, logit_bias=None, negative_only=False):
      B = image_features.shape[0]
      N = text_features.shape[0]
      chunk_size = min(self.chunk_size, B)

      total_loss = torch.zeros((), device=image_features.device, dtype=torch.float32)

      for i in range(0, B, chunk_size):
          end_i = min(i + chunk_size, B)
          img_chunk = image_features[i:end_i]
          logits = self.get_logits(img_chunk, text_features, logit_scale, logit_bias)

          # Treat every pair as negative:
          # -logsigmoid(-logits) == softplus(logits)
          chunk_loss = F.softplus(logits).sum()

          if not negative_only:
              # Replace local positives with positive-pair loss:
              # softplus(-x) - softplus(x) == -x
              num_pos = max(0, min(end_i, N) - i)
              if num_pos > 0:
                  rows = torch.arange(num_pos, device=logits.device)
                  pos_logits = logits[rows, i + rows]
                  chunk_loss = chunk_loss - pos_logits.sum()

          total_loss = total_loss + chunk_loss

      return total_loss / B
```

rwightman:
Reasoning provided, 

The current loss is:

  -F.logsigmoid(labels * logits).sum()

  For a negative pair, labels = -1, so:

  -F.logsigmoid(-logit)

  That is exactly:

  F.softplus(logit)

  For a positive pair, labels = +1, so:

  -F.logsigmoid(logit)

  That is exactly:

  F.softplus(-logit)

  Since almost every entry is negative, we can compute the whole chunk as if every pair were negative:

  loss = F.softplus(logits).sum()

  Then fix only the diagonal positives. For each positive logit x, we currently counted:

  softplus(x)

  but we need:

  softplus(-x)

  So add the correction:

  loss += softplus(-x) - softplus(x)

  Using the identity:

  softplus(-x) - softplus(x) == -x

  that correction can be simplified to:

  loss -= x

  So no label matrix is needed. The labels are just encoding “all negative except diagonal positive,” and that structure can be represented directly with the default negative loss plus a diagonal correction.

avocardio:
Nice — applied in 759f36f. Parity test still passes.

rwightman:
@avocardio thanks, merging now

Issue #1121, opened 2025-10-30
Evaluation error RuntimeError: ProcessGroupWrapper: Monitored Barrier encountered error running collective


samirchar:
I am training OpenClip on a distributed environment with 4 nodes and 8 GPUs per node. I set epochs to 40 and after training finishes with the last epoch (epoch 39) I am getting an error CUDA. Notice that it trains until the last epoch fine, including the evaluation after each epoch, but when it enters the last evaluation it fails. Here is the traceback below. Shouldn't the model be unwrapped of DDP for evaluate if only running eval on master? I think that while the last evaluation is running, the other ranks are finishing and DDP tries to sync and fails.

2025-10-30,14:14:29 | INFO | Start epoch 39
2025-10-30,14:14:32 | INFO | Train Epoch: 39 [   8192/2916352 (0%)] Data (t): 2.483 Batch (t): 3.019, 2713.79/s, 84.8059/s/gpu LR: 0.000002 Logit Scale: 88.598 Contrastive_loss: 0.046415 (0.046415) Loss: 0.046415 (0.046415)
2025-10-30,14:15:42 | INFO | Train Epoch: 39 [ 827392/2916352 (28%)] Data (t): 0.085 Batch (t): 0.699, 11850.7/s, 370.335/s/gpu LR: 0.000001 Logit Scale: 88.607 Contrastive_loss: 0.059184 (0.052800) Loss: 0.059184 (0.052800)
2025-10-30,14:16:51 | INFO | Train Epoch: 39 [1646592/2916352 (56%)] Data (t): 0.085 Batch (t): 0.699, 11751.0/s, 367.218/s/gpu LR: 0.000000 Logit Scale: 88.611 Contrastive_loss: 0.048979 (0.051526) Loss: 0.048979 (0.051526)
2025-10-30,14:18:01 | INFO | Train Epoch: 39 [2465792/2916352 (85%)] Data (t): 0.083 Batch (t): 0.697, 12237.6/s, 382.423/s/gpu LR: 0.000000 Logit Scale: 88.612 Contrastive_loss: 0.049657 (0.051059) Loss: 0.049657 (0.051059)
[...]

PR #1155, opened 2026-04-17
Hoist DDP unwrap to top of evaluate() so zero_shot_eval runs on bare module

### LINKED ISSUE #1152: evaluate(): DDP unwrap is nested under 'val' branch; zero_shot_eval still wrapped
## Problem

PR #1134 (fixes #1121) added `model = model.module` to avoid a DDP deadlock during non-zero-shot eval. The unwrap is placed *inside* the `if 'val' in data and ...` branch of `evaluate()`, after `zero_shot_eval(model, ...)` has already been called:

https://github.com/mlfoundations/open_clip/blob/main/src/open_clip_train/train.py#L251-L267

```python
def evaluate(model, data, epoch, args, tb_writer=None, tokenizer=None):
    metrics = {}
    if not is_master(args):
        return metrics
    device = torch.device(args.device)
    model.eval()

    zero_shot_metrics = zero_shot_eval(model, data, epoch, args, tokenizer=tokenizer)  # still DDP-wrapped
    metrics.update(zero_shot_metrics)

    ...
    if 'val' in data and (args.val_frequency and ((epoch % args.val_frequency) == 0 or epoch == args.epochs)):
        # unwrap DDP for single process eval
        if args.distributed and not args.horovod:
            model = model.module                            # scoped only here
```

Because rank > 0 exits via `if not is_master(args): return metrics` and then waits at the `torch.distributed.barrier()` added in `main.py` by #1134, rank 0 is inside `zero_shot_eval` still calling forward on a DDP-wrapped model. If that forward triggers the DDP reducer, rank 0 waits for ranks that are already at the barrier — the same symptom #1121 reported, shifted to the zero-shot path. The subsequent `imagenet-val` / `imagenet-v2` blocks further down in `evaluate()` also still run on the wrapped model.

## Proposed fix

Move the unwrap to the top of `evaluate()` so every eval path runs on the bare module:

```diff
 def evaluate(model, data, epoch, args, tb_writer=None, tokenizer=None):
     metrics = {}
     if not is_master(args):
         return metrics
     device = torch.device(args.device)
     model.eval()
+    if args.distributed and not args.horovod and hasattr(model, "module"):
+        model = model.module

     zero_shot_metrics = zero_shot_eval(model, data, epoch, args, tokenizer=tokenizer)
     metrics.update(zero_shot_metrics)

     autocast = get_autocast(args.precision, device_type=device.type)
     input_dtype = get_input_dtype(args.precision)

     if 'val' in data and (args.val_frequency and ((epoch % args.val_frequency) == 0 or epoch == args.epochs)):
-        # unwrap DDP for single process eval
-        if args.distributed and not args.horovod:
-            model = model.module
         dataloader = data['val'].dataloader
```

Follow-up to #1134. I will open the PR tied to this issue.

shaun0927:
Follow-up to #1134. The `model = model.module` unwrap added in that PR is nested inside the `if 'val' in data ...` branch, so `zero_shot_eval` (and the `imagenet-val` / `imagenet-v2` blocks) still run on a DDP-wrapped model on rank 0. With `if not is_master(args): return metrics` exiting rank > 0 early, any DDP reducer activity inside `zero_shot_eval` can deadlock against the barrier that #1134 added in `main.py` — the same symptom #1121 reported, shifted to the zero-shot path.

## Fix

Unwrap once at the top of `evaluate()`, right after the master guard. Every eval path then runs on the bare module.

Fixes #1152

### DIFF
--- a/src/open_clip_train/train.py
+++ b/src/open_clip_train/train.py
@@ -254,6 +254,10 @@ def evaluate(model, data, epoch, args, tb_writer=None, tokenizer=None):
         return metrics
     device = torch.device(args.device)
     model.eval()
+    # unwrap DDP once here so zero_shot_eval and the imagenet-val / imagenet-v2
+    # paths below all run on the bare module on rank 0 (completes #1134).
+    if args.distributed and not args.horovod and hasattr(model, "module"):
+        model = model.module
 
     zero_shot_metrics = zero_shot_eval(model, data, epoch, args, tokenizer=tokenizer)
     metrics.update(zero_shot_metrics)
@@ -262,9 +266,6 @@ def evaluate(model, data, epoch, args, tb_writer=None, tokenizer=None):
     input_dtype = get_input_dtype(args.precision)
 
     if 'val' in data and (args.val_frequency and ((epoch % args.val_frequency) == 0 or epoch == args.epochs)):
-        # unwrap DDP for single process eval
-        if args.distributed and not args.horovod:
-            model = model.module
         dataloader = data['val'].dataloader
         num_samples = 0
         samples_per_val = dataloader.num_samples
--- a/src/open_clip_train/zero_shot.py
+++ b/src/open_clip_train/zero_shot.py
@@ -49,7 +49,7 @@ def zero_shot_eval(model, data, epoch, args, tokenizer=None):
         return {}
     if (epoch % args.zeroshot_frequency) != 0 and epoch != args.epochs:
         return {}
-    if args.distributed and not args.horovod:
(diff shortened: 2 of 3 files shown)


shaun0927:
Addressed the zero-shot DDP regression in `d726c8a`.

- made `zero_shot_eval()` unwrap only when `model.module` actually exists
- kept the earlier `evaluate()` hoist intact
- added focused regression coverage for both already-unwrapped and wrapped distributed inputs

Validated locally:
- `cd /Users/jh0927/Workspace/open_clip-pr1155-followup && . .venv/bin/activate && python -m pytest tests/test_zero_shot_eval.py -q`

PR #1168, opened 2026-06-01
NaFlexGenLIP experiment

rwightman:
Had a fun session creating an implementation on GenLIP ( https://github.com/YanFangCS/GenLIP) that uses naflex style image resize and batch creation instead of navit style full sequence packing.
* 'NaFlexGenLIP' model that's a hybrid of the original GenLIP model w/ NaFlexViT style input compatibility
* OpenCLIP NaFLex data pipeline allocates space for caption tokens in budgeting
* Dataset analyzer script to suggest good batch/seq len constraints to get reasonable padding/truncation based on caption stats
* 'zero-shot' image classification script (rather experimental) based on template string probs
* Sequence length bucketing impl to reduce padding overhead 
* Allow caption/text extraction from json fields


bryant1410:
Ooc, will you try to reproduce trianing the model?

rwightman:
@bryant1410  I trained the b16 on cc12m locally, achieved linear probe of ~38% at seq len 256 and 39.75 at 384.... so I consider the pieces working

PR #1181, opened 2026-06-23
Support extra timm transform parameters

igormp:
timm has had support for the `color_jitter_prob` and `gray_scale_prob` parameters since its version v0.9.16[^1], which were introduced by https://github.com/huggingface/pytorch-image-models/pull/2064

Given how its dependency version was bumped by #1097 (in this commit: https://github.com/mlfoundations/open_clip/pull/1097/changes/6641590880a7a540eec94ec4bb8203635df0dc99), it should be safe to use those parameters from the `AugmentationCfg` class without popping them out, while still keeping sane defaults in order to maintain backwards compatibility.

Also taking the chance to move `hflip` to the config as well, allowing one to adjust it now, and adding `vflip` while we are at it.

I wonder how much of timm's `create_transform` arguments should be surfaced here, and if a bigger refactor of this module should take place, given that the dependency on timm doesn't look to be optional anymore since https://github.com/mlfoundations/open_clip/pull/357, so all code paths that have it as optional could likely be dropped.

[^1]: https://github.com/huggingface/pytorch-image-models/releases/tag/v0.9.16

### DIFF
--- a/src/open_clip/transform.py
+++ b/src/open_clip/transform.py
@@ -411,6 +411,8 @@ def image_transform(
             else:
                 input_size = (3, image_size, image_size)
 
+            aug_cfg_dict.setdefault('color_jitter', None)  # disable by default
+
             timm_transform_kwargs = dict(
                 input_size=input_size,
                 is_training=True,
--- a/src/open_clip/transform.py
+++ b/src/open_clip/transform.py
@@ -61,6 +61,8 @@ def merge_preprocess_kwargs(base: PreprocessCfg, **kwargs):
 
 @dataclass
 class AugmentationCfg:
+    hflip: float = 0.0
+    vflip: float = 0.0
     scale: Tuple[float, float] = (0.9, 1.0)
     ratio: Optional[Tuple[float, float]] = None
     color_jitter: Optional[Union[float, Tuple[float, float, float], Tuple[float, float, float, float]]] = None
@@ -409,15 +411,9 @@ def image_transform(
             else:
                 input_size = (3, image_size, image_size)
 
-            aug_cfg_dict.setdefault('color_jitter', None)  # disable by default
-            # drop extra non-timm items
-            aug_cfg_dict.pop('color_jitter_prob', None)
-            aug_cfg_dict.pop('gray_scale_prob', None)
-
             timm_transform_kwargs = dict(
                 input_size=input_size,
                 is_training=True,
-                hflip=0.,
                 mean=mean,
                 std=std,
                 re_mode='pixel',
--- a/src/open_clip/transform.py
+++ b/src/open_clip/transform.py
@@ -1,7 +1,7 @@
 import numbers
 import random
 import warnings
-from dataclasses import dataclass, asdict
+from dataclasses import dataclass, asdict, InitVar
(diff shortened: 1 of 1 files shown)


rwightman:
@igormp to avoid changing defaults or breaking things I think this is needed?

```
  aug_cfg_dict.setdefault("color_jitter", None)
  if "gray_scale_prob" in aug_cfg_dict:
      aug_cfg_dict["grayscale_prob"] = aug_cfg_dict.pop("gray_scale_prob")
```

igormp:
@rwightman thanks for the catch on `color_jitter`, I totally glossed over that the `aug_cfg_dict` was dropping `None` earlier.

I also clearly had not noticed the slightly different spelling between `grayscale_prob` and `gray_scale_prob`. I ended up trying to make the dataclass support both parameters while trying to maintain backwards compatibility (and updated the parameter name throughout the code base).
It's a bit more complex than simply popping from a dict, but I feel like it's more flexible, what do you think?

Rebased against `main`. Let me know if you'd like me to clean up my commits as well.

rwightman:
@igormp I guess we can do the bwd compat thing to bring the names into alignment, you're likely to get a warning on hflip/vflip there though, so either add hflip/vflip support to the non timm pipeline or do more default to None / set default combos ...

ErenAta16 (changes requested):
Traced how `aug_cfg_dict` actually gets used further down in `image_transform` (line ~413: `aug_cfg_dict = {k: v for k, v in asdict(aug_cfg).items() if v is not None}`, then merged in at line ~433 via `timm_transform_kwargs = dict(..., **aug_cfg_dict)`, then `create_transform(**timm_transform_kwargs)`). Removing the hardcoded `hflip=0.,` is correct and necessary now that `hflip` is a real `aug_cfg_dict` key, that part looks right, otherwise you'd get "got multiple values for keyword argument 'hflip'" once both are present.

But the same removal also dropped this:

```python
# drop extra non-timm items
aug_cfg_dict.pop('color_jitter_prob', None)
aug_cfg_dict.pop('gray_scale_prob', None)
```

`color_jitter_prob` and `grayscale_prob` (the renamed field) are consumed by the non-timm branch only, the block near the bottom that does `gray_scale(aug_cfg.grayscale_prob)` and `color_jitter(*aug_cfg.color_jitter, p=aug_cfg.color_jitter_prob)` is the `use_timm=False` path. They were never timm `create_transform` parameters, that's exactly what the removed comment says. With the pop gone, if a caller sets `use_timm=True` together with a non-None `color_jitter_prob` or `grayscale_prob` (both realistic, e.g. someone reusing an existing `AugmentationCfg` while switching to timm transforms), `aug_cfg_dict` still carries those two keys into `**aug_cfg_dict`, and `create_transform()` doesn't accept them, so this raises `TypeError: create_transform() got an unexpected keyword argument 'color_jitter_prob'` instead of silently ignoring them like before.

Looks like the fix should keep those two specific pops (updating the second one to `'grayscale_prob'`) while still letting `hflip`/`vflip` (and anything else that's a genuine timm parameter) through. Everything else, the `InitVar` back-compat shim for the rename and the `hflip`/`vflip` additions themselves, looks correct.

PR #1189, opened 2026-07-06
Fix NaFlex worker-padded sample count after schedule shuffle

chenghuichen:
When NaFlex schedules are shuffled before worker padding, the repeated padding batch can differ by epoch. 
num_samples_for_workers() computes metadata from the unshuffled padded schedule, so dataloader.num_samples can disagree with the actual number of rows yielded.

Example:
canonical: [(8, 4), (4, 8), (4, 5)] -> 17 samples
epoch 0 shuffled+worker-padded actual sum: 22
advertised num_samples_for_workers(4): 21

Fix: pad the schedule before epoch shuffle, so worker-padded batch count and sample count stay stable across epochs.

### DIFF
--- a/src/open_clip_train/naflex_data.py
+++ b/src/open_clip_train/naflex_data.py
@@ -641,9 +641,9 @@ class NaFlexBatchScheduler:
 
     def epoch_schedule(self, epoch: int, num_workers: int = 1) -> List[Tuple[int, int]]:
         schedule = list(self._canonical_batch_schedule)
+        schedule = self.pad_schedule_for_workers(schedule, max(1, num_workers))
         if self.shuffle:
             random.Random(self.seed + epoch).shuffle(schedule)
-        schedule = self.pad_schedule_for_workers(schedule, max(1, num_workers))
         return schedule
 
     @staticmethod
--- a/src/open_clip_train/naflex_data.py
+++ b/src/open_clip_train/naflex_data.py
@@ -639,9 +639,14 @@ class NaFlexBatchScheduler:
     def __len__(self) -> int:
         return self.num_batches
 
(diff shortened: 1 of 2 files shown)


rwightman:
@chenghuichen thanks, def a valid issue,  fix looks like it should work. I did a review and this was proposed as more robust soln (avoid possible mismatch), but a bit of list iteration overhead each epoch...  thoughts?

```python
  def _worker_padded_schedule(self, num_workers: int) -> List[Tuple[int, int]]:
      return self.pad_schedule_for_workers(
          list(self._canonical_batch_schedule),
          max(1, num_workers),
      )

  def epoch_schedule(self, epoch: int, num_workers: int = 1) -> List[Tuple[int, int]]:
      schedule = self._worker_padded_schedule(num_workers)
      if self.shuffle:
          random.Random(self.seed + epoch).shuffle(schedule)
      return schedule

  def num_batches_for_workers(self, num_workers: int) -> int:
      return len(self._worker_padded_schedule(num_workers))

  def num_samples_for_workers(self, num_workers: int) -> int:
      schedule = self._worker_padded_schedule(num_workers)
      num_samples_per_rank = sum(batch_size for _, batch_size in schedule)
      if self.distributed:
          return num_samples_per_rank * self.world_size
      return num_samples_per_rank
```

chenghuichen:
@rwightman Thanks, agreed that extracting a helper is a good idea.

On the overhead: since num_workers is provided at runtime, I think there are two options: follow your suggested helper approach, which rebuilds the worker-padded schedule roughly once per epoch per worker, or add a cache keyed by num_workers. I’d prefer the helper-only version here. The cost should be negligible since the schedule is just metadata, and it keeps the implementation clean.

I’ve updated the PR accordingly.

PR #1214, opened 2026-09-12
Bound SigLIP chunked training memory with activation recomputation

trueoneplusone:
# Bound SigLIP chunked training memory with activation recomputation

With `SigLipLoss(chunk_size > 0)`, the forward loop creates small logit blocks, but autograd retains the logits from every block for `softplus` backward. The saved pairwise activations therefore still grow quadratically with batch size. This affects the memory-saving training use case introduced in #1145 and is related to #942.

This change checkpoints each complete pairwise-loss block with non-reentrant activation checkpointing. Backward recomputes the block's projection, fp32 reduction, and positive-pair correction instead of retaining all blocks' logits. Chunk offsets are explicit checkpoint inputs, and distributed collectives stay outside the recomputed function. The default unchunked path is unchanged. Frozen towers, trainable scale/bias, optional bias, and evaluation without gradients retain their behavior.

The tradeoff is additional backward computation in the already opt-in chunked path. Feature and distributed communication storage are not eliminated, and this does not claim to solve every distributed scaling issue in #942.

## Validation

- Added a regression that counts actual, deduplicated autograd-saved storage, excluding storage shared with the original inputs. Before the change, all four memory cases fail, including cases where only logit scale or bias is trainable; the other 90 correctness cases pass.
- Added loss and all-input gradient parity across chunk sizes, uneven/rectangular feature batches, negative-only blocks, frozen towers, optional bias, CPU bf16 autocast, and no-grad evaluation.
- Added differentiable remote-block tests for gather/reduce/shift/bidir and world sizes 2/3/4, checking loss/gradient parity and that backward does not replay communication. These supply remote features directly and do not replace real multi-process integration testing.
- `160 passed, 5 deselected` for the five loss test modules listed below. The five deselected cases instantiate full ViT caption models; the remaining caption-loss unit tests run.
- Small CUDA fp32/fp16/bf16 examples match the previous chunked implementation's loss and image/text/scale/bias gradients. `torch.compile(fullgraph=True, backend="eager")` forward and gradient smoke also passes.
- `git diff --check` passes.

```bash
python -m pytest tests/test_siglip_chunked_loss.py tests/test_siglip_loss_precision.py tests/test_loss.py tests/test_factory_loss.py tests/test_fused_caption_loss.py -k 'not test_fused_caption_loss_matches_legacy' -q
```

CPU saved pairwise storage before/after (fp32, embedding dimension 32):

| Batch | Chunk size | Before | After |
| --- | --- | --- | --- |
| 256 | 32 / 128 | 256 KiB | 0 additional saved pairwise storage |
| 512 | 32 / 128 | 1 MiB | 0 additional saved pairwise storage |
| 1024 | 32 / 128 | 4 MiB | 0 additional saved pairwise storage |

Checkpoint inputs still retain the original feature storage; zero in this table is not zero training memory.

Small CUDA forward + backward peak allocated-memory increments, measured above preallocated inputs after warmup with `torch.cuda.max_memory_allocated()` (batch 2048, embedding dimension 64, chunk size 64, RTX 5090, PyTorch 2.10.0+cu130):

| Projection dtype | Previous chunked implementation | This change |
| --- | --- | --- |
| fp32 | 18.016 MiB | 2.518 MiB |
| fp16 autocast | 25.524 MiB | 2.760 MiB |
| bf16 autocast | 25.524 MiB | 2.760 MiB |

These are isolated loss measurements, not whole-model training benchmarks or a throughput claim. A real three-process Gloo check could not reach the loss because this Windows build fails during process-group initialization with `makeDeviceForHostname(): unsupported gloo device`; distributed multi-process validation remains a limitation.

Developed with AI assistance; the regression failures, test results, and small CUDA measurements above were run locally.

### DIFF
--- a/src/open_clip/loss.py
+++ b/src/open_clip/loss.py
@@ -436,7 +436,10 @@ class SigLipLoss(nn.Module):
     def _chunked_loss(self, image_features, text_features, logit_scale, logit_bias=None, negative_only=False):
         """Memory-efficient loss that chunks the logit computation.
 
-        Peak memory: O(chunk_size * N) instead of O(B * N).
+        Pairwise activation memory: O(chunk_size * N) instead of O(B * N).
+        Under autograd, each chunk is checkpointed and recomputed in backward
+        so earlier chunks' logits do not remain live until the final reduction.
+        This trades additional computation for memory; feature storage is unchanged.
         Useful when per-device batch is large (e.g. B > 4096).
 
         Uses the identities -logsigmoid(-x) = softplus(x) and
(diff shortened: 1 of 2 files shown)


trueoneplusone:
Hi maintainers, following up on this SigLIP activation-recomputation PR. CI is green, and the change is scoped to the opt-in chunked-loss path (`chunk_size > 0`), with forward/gradient parity and memory tests included. I would appreciate feedback on whether non-reentrant checkpointing at the pairwise-loss block boundary fits the project's preferred design. Happy to adjust the implementation or testing if needed.

